PR #214, opened 2019-08-06
Progress Bar 

mrturkmen06:
Progress bar is added to client side, so when a user want to create an event, the status of VMs will be shown on client terminal with percentages. 
[Demonstration](https://asciinema.org/a/ajXLxG6o1Rxx4ZXCEBTK6nkPs?t=30)
[asciicast](https://asciinema.org/a/ajXLxG6o1Rxx4ZXCEBTK6nkPs)


kdhageman:
@mrturkmen06 Can you try to rebase on the 1.5.0 branch (which should equal the master branch as of now)? It's a bit difficult to review >1500 changed lines of which most are from previous commits.

app/client/cli/event.go
```diff
@@ -51,9 +51,7 @@ func (c *Client) CmdEventCreate() *cobra.Command {
 		Example: `hkn event create esboot -name "ES Bootcamp" -a 5 -c 30 -e scan,sql,hb -f kali`,
 		Args:    cobra.MinimumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
-			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
-			defer cancel()
-
+			ctx := context.Background()
```
kdhageman: This completely removes any timing out, do we want that?

app/client/cli/event.go
```diff
@@ -63,22 +61,20 @@ func (c *Client) CmdEventCreate() *cobra.Command {
...
+				statusReport, err := stream.Recv()
 				if err == io.EOF {
 					break
 				}
-
 				if err != nil {
 					PrintError(err)
 					return
 				}
+				fmt.Println(statusReport.Status)
```
kdhageman: I would suggest to have the progress bar on the client side, rather than on the daemon side. This bar is then incremented every time that a message is retrieved from the stream

app/client/cli/root.go
```diff
@@ -84,7 +84,7 @@ func versionCheckInterceptor(ctx context.Context, method string, req interface{}
 
 	ok, err := isClientVersionLessThan(daemonVersion)
 	if err != nil {
-		return fmt.Errorf("Unable to read daemon's version: %s", err)
+		return nil
```
kdhageman: Why ignore the error?

daemon/daemon.go
```diff
@@ -60,7 +59,8 @@ var (
 )
 
 const (
-	mngtPort = ":5454"
+	MAINCONFIG_PATH = "/scratch/configs/"
```
kdhageman: Can be removed?

daemon/daemon.go
```diff
@@ -476,6 +479,8 @@ func (d *daemon) createEventFromEventFile(ef store.EventFile) error {
 	return d.createEvent(ev)
 }
 
+
+//todo: DOES NOT CREATE EVENT, IT READS CONFIGURATION FILE FROM HOST AND CALL CREATEEVENT FUNCITON WITHHIN IT
```
kdhageman: So then it *does* create an event from a config file, doesn't it?

daemon/daemon.go
```diff
@@ -508,7 +515,32 @@ func (d *daemon) createEvent(ev event.Event) error {
...
+
 func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEventServer) error {
+	// Since there are more than one goroutines
+	// there should be synchronization of them
+	// to be able to send data back to client
+	// goroutines should be waited
+	var wg sync.WaitGroup
+	wg.Add(2)
+	var m sync.Mutex
+	m.Lock()
```
kdhageman: What is the purpose of the lock when it is unlocked immediately below (line 535)?

daemon/daemon.go
```diff
@@ -559,7 +592,36 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
...
+		Int("Capacity", conf_.Capacity).
+		Strs("Frontends", frontendNames).
+		Msg("Creating event")
+
+	// this goroutine will initialize CTFd module
+	// while other one getting information from
+	// lab.Status
+	go func() {
+		defer wg.Done()
+		m.Lock()
```
kdhageman: This is never unlocked?

event/event.go
```diff
@@ -11,8 +11,19 @@ import (
...
+	"github.com/aau-network-security/haaukins/lab"
+	"github.com/aau-network-security/haaukins/store"
+	"github.com/aau-network-security/haaukins/svcs/ctfd"
+	"github.com/aau-network-security/haaukins/svcs/guacamole"
+	"github.com/aau-network-security/haaukins/virtual/docker"
+	"github.com/aau-network-security/haaukins/virtual/vbox"
+	"github.com/rs/zerolog/log"
 	"io"
 	"sync"
+<<<<<<< HEAD
```
kdhageman: These `<<<<`, `=====` and `>>>>>>` must be removed, as they are git specific

lab/hub.go
```diff
@@ -7,16 +7,18 @@ package lab
...
 	"sync/atomic"
-
 	"github.com/aau-network-security/haaukins/store"
 	"github.com/aau-network-security/haaukins/virtual/vbox"
 	"github.com/rs/zerolog/log"
 )
 
+
 var (
+	Status   = make(chan string,5)
```
kdhageman: This is really dangerous. We have a shared variable that is shared among all clients, hub, events, etc. Imagine a scenario where two clients are creating an event simultaneously, then both will read from the same channel and get the wrong information.

kdhageman: I would suggest passing a `context` to the `NewHub` function. When the `daemon` receives a `createEvent` message, it attaches a some kind of logging object to this context. Then the in `NewHub`, this logging object is extracted from the context.

For inspiration, you can look at [this link](https://blog.gopheracademy.com/advent-2016/context-logging/) where they do something similar

lab/lab.go
```diff
@@ -11,7 +11,10 @@ import (
 
 	"io"
 	"sync"
+<<<<<<< HEAD
```
kdhageman: As mentioned before, should be removed

lab/lab.go
```diff
@@ -42,7 +45,9 @@ type LabHost interface {
 	NewLab(context.Context, vbox.Library, Config) (Lab, error)
 }
 
-type labHost struct{}
+type labHost struct{
```
kdhageman: Empty struct?

svcs/guacamole/guacamole.go
```diff
@@ -134,7 +154,8 @@ func (guac *guacamole) create(ctx context.Context) error {
 		},
 	})
 
-	mysqlPass := uuid.New().String()
+	mysqlPass := "deneme"
```
kdhageman: I'm not really for hardcoding passwords

users.yml
```diff
@@ -0,0 +1,5 @@
+users:
```
kdhageman: Maybe add `users.yml` to `.gitignore`, so we don't accidentally commit configuration files (they don't necessarily belong on git)

virtual/vbox/vbox.go
```diff
@@ -55,9 +55,22 @@ func (err *VBoxErr) Error() string {
 type VM interface {
 	virtual.Instance
 	Snapshot(string) error
-	LinkedClone(context.Context, string, ...VMOpt) (VM, error)
+	LinkedClone(context.Context, string, ...VMOpt) (VM, error) // what is LinkedClone ?
```
kdhageman: See [this link](https://docs.oracle.com/cd/E97728_01/E97727/html/vboxmanage-clonevm.html) of VirtualBox documentation. A linked clone is a clone of an existing VM whose disk is defined by its diff against the original VM (if I recall correctly). This is opposed to a full clone, whose disk is a full copy of the original one. So disk (and speed)wise, it is much more efficient to create linked clones than full clones.

app/client/cli/event.go
mrturkmen06: It will be canceled out when VMs are started, when all VM's are properly started, the channel will be closed so there will be EOF, then it will automatically closed.

lab/hub.go
kdhageman: And to add to this suggestion, you could implement this vaguely described *some kind of  logging object* a new interface that has a `Stream`, `Increment` or `Log` function or something.

daemon/daemon.go
mrturkmen06: It were but I have modified it to send some stream of data back to client.

daemon/daemon.go
mrturkmen06: removed

app/client/cli/root.go
mrturkmen06: I think, it is left when I was trying to setup local Haaukins platform, I just tried to see what happen when I ignore it. I should rewrite it down.

app/client/cli/event.go
mrturkmen06: Yeah, but what will be the difference ? Same procedure will applied again.

event/event.go
mrturkmen06: Yeah.

svcs/guacamole/guacamole.go
mrturkmen06: I have used it while I was trying to understand how data stores. I will re-add randomised one of course.

users.yml
mrturkmen06: 👍🏻

app/client/cli/event.go
kdhageman: One difference will be that you will not see this stack of progress bars (like your example gif), but a single one that updates dynamically (which I prefer). I guess it's also a matter of responsibility; the daemon should not really be concerned with how the client displays the information that it sends.

app/client/cli/event.go
kdhageman: What if the daemon hangs indefinitely (I am playing the devil's advocate here 😉)? Should the client also hang indefinitely? I am fine with changing it, but I am just making sure that you know of the implication of changing this.

mrturkmen06: However, if some VMs are NOT initialised correctly in given timeout period, then we might have other issues as well. If there is a problem on Daemon side, it will also cancel out on client side.

daemon/daemon.go
mrturkmen06: Yeah, it is left from other tests. removed.

daemon/daemon.go
mrturkmen06: removed.

lab/lab.go
mrturkmen06: done

svcs/guacamole/guacamole.go
mrturkmen06: replaced.

app/client/cli/event.go
kdhageman: 👍

virtual/vbox/vbox.go
mrturkmen06: 👍🏻 I will check it.

kdhageman: So there exists always a "master" or "parent" VM, from which we create clones. You can see this on `sec02` when you run the following:
```
ntpd@sec02:~$ vboxmanage list vms
...
"kali{ecc3ea71}" {c8a577db-b6e7-4b9a-b79d-cd4fed8edc08}
...
"40541ce8f1e54c31940406462144e906" {6b2045aa-74ec-4005-a609-2da467add676}
"754a397f117d455cb3232a90685f1611" {a09c717f-2700-4969-92e6-d05ce662da8f}
"941ebe8faac142d68dccb9521de0d32e" {2e0bad87-126a-43be-878d-aa8f68f884ff}
"0868359c8e314aeb9b9a355657a79e64" {398ff831-e10e-4312-990b-c2b34c01a92f}
"1cc94f0b7650468ab7a42913e5e61e25" {1c1bf14a-03ce-4af4-b1dc-fa0a5d985d38}
...
```
Here, `kali{ecc3ea71}` is the "master" or "parent" VM, and all the other ones are linked clones of that one.

mrturkmen06:
Changes and required functions implemented, there is no stack of progress bar anymore and more importantly, there is no shared variable. 
<a href="https://asciinema.org/a/E0JMyfqu277sVDbGSLiGrfWXq" target="_blank"></a>

daemon/daemon.go
```diff
@@ -559,7 +571,34 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 		conf.Capacity = 10
 	}
 
-	return d.createEventFromConfig(conf)
+	loggerInstance := &GrpcLogger{resp: resp}
```
kdhageman: This code was taken from another function right? Then I would suggest to remove the other function to prevent duplicate code

logs/audit.log
```diff
@@ -0,0 +1,268 @@
+{"level":"info","username":"mrturkmen","time":"2019-07-22T08:49:22Z","message":"signup user"}
```
kdhageman: Prefer to remove this file from repo (I know it is already ignored by `.gitignore`, but this file still remains in PR).

If you do something like this, you keep the file locally on your machine. but remove it from git (need to commit afterwards):
`git rm --cached logs/audit.log`

users.yml
kdhageman: Same as `audit.log`

daemon/daemon.go
mrturkmen06: Yeah, I thought to remove it as your suggestion however there are some usage of createEvent function internally, it is used on other points for this reason I did not remove it. I tried to express usage of it on screenshot below. Since the functions which starts with Capital and small letters differ in go,  I prefer to keep it to prevent further errors. Do you think that is it ok ?
Screenshot 2019-08-12 at 13 19 28

mrturkmen06: This part of the code creates a lot of confusion, refactoring process might be useful especially for this piece of code.

logs/audit.log
mrturkmen06: ok

PR #224, opened 2019-08-13
docker image versions added

mrturkmen06:


svcs/dhcp/dhcp.go
```diff
@@ -47,9 +47,8 @@ func New(format func(n int) string) (*Server, error) {
 	if err != nil {
 		return nil, err
 	}
-
 	cont := docker.NewContainer(docker.ContainerConfig{
-		Image: "networkboot/dhcpd",
+		Image: "networkboot/dhcpd",  // no need to add tag since it is not updated for 5 months.
```
kdhageman: Just be aware that this is not a guarantee that it doesn't change in the future.

kdhageman: I'm still merging though :)

mrturkmen06: :) no need to worry, if it does change then we will notify it, additionally there is no tag other than latest

PR #226, opened 2019-08-14
Ram issue

kdhageman:


kdhageman:
@mrturkmen06 ready for review?

mrturkmen06:
@kdhageman, The issue is fixed, so VMs created with defined memory size in configuration file, since this was draft pull request, I have changed and committed the changes to branch and mark it as ready to review.

mrturkmen06:
To make merge, I do not have authority to do that.

virtual/vbox/vbox_test.go
```diff
@@ -83,3 +86,57 @@ func TestVmBase(t *testing.T) {
...
+	cs := "d41d8cd98f00b204e9800998ecf8427e"
+	vm := vbox.NewVMWithSum("haaukins.ova", "haaukins", cs)
+	vm.Create(ctx)
+	vm.Snapshot("test_haaukins")
+
+	linkedCloneVM,err := vm.LinkedClone(ctx,"test_haaukins",vbox.SetRAM(uint(memorysize)),vbox.SetCPU(uint(2)))
+	if err != nil {
+		t.Fatalf("Linked clone could not created %s ", err)
+	}
+
```
kdhageman: Be careful with closing the VMs. For instance, if the next function call fails, the `linkedCloneVm` will never close, resulting in a dangling VM on the system that is running the test. So, I'd recommend using `defer` statements instead. That will ensure that a certain function will be called, regardless of test failure.

mrturkmen06: 👍🏻 Thank you for the feedback :) Updating...

store/frontend_test.go
```diff
@@ -82,3 +83,25 @@ func TestSetMemoryMBAndCpu(t *testing.T) {
 		})
 	}
 }
+
+func TestNewFrontendsFile(t *testing.T) {
+	// since there is no dummy yml file on git repo, skipping test on Travis can be reasonable...
+	tst.SkipCI(t)
+	definedMemory :=uint(4096)
+	definedCPU := 2.0
+	frontendstore, err := NewFrontendsFile("test_ymls/frontend_test.yml")
```
kdhageman: Suggestion: change to a temporary file that you delete in the test?

mrturkmen06: 👍🏻

PR #228, opened 2019-08-16
Replace CTFd with internal Go [will be removed]

tpanum:
Trying to create a frontend service for events that will eventually replace CTFd.


tpanum:
I will move Amigo its own repo, this is currently just for backup.

Issue #188, opened 2018-11-23
White screen when pressing connect


DanielBritze:
When people register and press login, they cannot connect, they only receive a white screen in the new guac tab.
```
Event: | Date | Approx time | Team ID | Type | fix kode | virkede det? | System og browser
Aars HHX | 21/11/2018 | 9:39 | 2d49f2a4 | Guac connect virker ikke | Not avialable | Nej | Macbook air - chrome
Aars HHX | 21/11/2018 | 9:40 | 6c5797ab | Guac connect virker ikke | Not avialable | Nej | Macbook air - chrome
Aars HHX | 21/11/2018 | 9:46 | unkown | Guac connect virker ikke | Not avialable | Nej | Macbook - chrome
Aars HHX | 21/11/2018 | 9:45 | bed5a209 | Guac connect virker ikke | Not avialable | Nej | Macbook - safari
Aars HHX | 21/11/2018 | 9:45 | 5f32a235 | Guac connect virker ikke | Not avialable | Nej | Macbook - chrome
Aars HHX | 21/11/2018 | 9:50 | ef89a5e6 | Guac connect virker ikke | Not avialable | Nej | Macbook - chrome
Aars HTX | 21/11/2018 | 12:30 | -- | Guac connect virker ikke |   |   | acer- windows chrome
Aars HTX | 21/11/2018 | 12:30 | teamscm | Guac connect virker ikke |   |   | Asus - windows - chrome
```

mrturkman:
Hi, 

This might me delayed observation regarding to white screen, nevertheless I wanted to share what I found regarding to "white screen " issue. 

When a user connected to Haaukins platform over Google Chrome browser and click "Connect" button, they face "White screen" as you described. However, it does not happen in other browsers such as Safari & Firefox.  SS of the logs given below when an user tried to connect platform over Chrome. 
![Screenshot 2019-07-17 at 11 51 29](https://user-images.githubusercontent.com/13614433/61366147-50bc1480-a889-11e9-928d-ef2dcf57a890.png)

So, I think, the issue (if it does not happen in other browsers (Firefox, Safari) ) is related with the dependent library *(Guacamole), which enables users to connect Kali over browser by creating RDP.  All in all, there might be some misconfiguration within Go files or another bug, it will be discovered and solved. This is just an observation from log files and tests on other browsers, other inspections might take place regarding to white screen issue. 

Regards.

tpanum:
[...]

PR #230, opened 2019-08-19
Added queuing of teams when assigning labs

jbjjbjjbj:
This may solve the white screen issue described in #188. 

This solves the issue where a register request is discarded because of the lack of labs. 
Instead they will be added to a queue and will be assigned to a lab when they come available.

This has only been testet when registering two teams with a availability of one. 

TODO
- [x] Check if lab is assigned when pressing connect
- [x] Check if hub has reached it's capacity

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -115,6 +115,8 @@ type event struct {
 	dockerHost    docker.Host
 
 	closers []io.Closer
+
+	teamQueue chan *store.Team
 }
 
 func NewEvent(ctx context.Context, ef store.EventFile, hub lab.Hub) (Event, error) {
@@ -152,7 +154,12 @@ func NewEvent(ctx context.Context, ef store.EventFile, hub lab.Hub) (Event, erro
 		closers:       []io.Closer{ctf, guac, hub, keyLoggerPool},
 		dockerHost:    dockerHost,
 		keyLoggerPool: keyLoggerPool,
+		teamQueue:     make(chan *store.Team, conf.Capacity),
 	}
+
+	hub.AttachHook(func() error {
+		return ev.processQueue()
+	})
 
 	return ev, nil
 }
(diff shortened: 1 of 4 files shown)


event/event.go
```diff
@@ -152,8 +154,13 @@ func NewEvent(ctx context.Context, ef store.EventFile, hub lab.Hub) (Event, erro
 		closers:       []io.Closer{ctf, guac, hub, keyLoggerPool},
 		dockerHost:    dockerHost,
 		keyLoggerPool: keyLoggerPool,
+		teamQueue:     make(chan *store.Team, conf.Capacity),
 	}
 
+	hub.AttachHook(func() error {
+		return ev.processQueue()
```
jbjjbjjbj: This may have unwanted effects, as this error has nothing to do with adding and removing vm's.

tpanum:
Great work @jbjjbjjbj 

If you need any help or have questions don't hesitate to come by our office!

event/event.go
```diff
@@ -284,15 +291,12 @@ func (ev *event) AssignLab(t *store.Team) error {
 
 func (ev *event) Handler() http.Handler {
 	reghook := func(t *store.Team) error {
-		if ev.labhub.Available() == 0 {
-			return lab.CouldNotFindLabErr
-		}
-
-		if err := ev.AssignLab(t); err != nil {
-			return err
+		if len(ev.labhub.GetLabs())+len(ev.teamQueue) >= ev.GetConfig().Capacity {
```
jbjjbjjbj: Kind of ugly. Is there a better way?

svcs/guacamole/intercept.go
```diff
@@ -6,14 +6,228 @@ package guacamole
...
 import (
 	"errors"
-	"fmt"
 	"net/http"
 	"sync"
 
 	"github.com/aau-network-security/haaukins/store"
 	"github.com/rs/zerolog/log"
 )
+const waitingHTMLTemplate = `
```
mrturkmen06: A fancy waiting UI added as constant, I think, it could be useful to define someplace where all html & css stuff resides. Seems dirty here however it is fancy on browser :)

event/event.go
mrturkmen06: There could be however that has no priority for now. So, lets leave it and merge the branch for now.

PR #232, opened 2019-08-20
Close function is modified, VMs can be shutdown properly.

mrturkmen06:
The issue regarding to VMs (which do not shutdown properly) fixed, so whenever an event stopped, VMs will be destroyed which provides more resources than before.

### DIFF
--- a/lab/lab.go
+++ b/lab/lab.go
@@ -209,25 +209,31 @@ func (l *lab) Restart(ctx context.Context) error {
 
 func (l *lab) Close() error{
 	var wg sync.WaitGroup
+
 	 for _, lab  := range l.frontends {
 	 	wg.Add(1)
-	 	go func (){
+	 	go func (vm vbox.VM){
 	 		// closing VMs....
 	 		defer wg.Done()
-			if err := lab.vm.Close(); err!=nil{
+			if err := vm.Close(); err!=nil{
 				log.Error().Msgf("Error on Close function in lab.go %s",err)
 			}
-		}()
+		}(lab.vm)
 	 }
-	 go func() {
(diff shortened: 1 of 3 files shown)


lab/lab.go
```diff
@@ -208,24 +207,26 @@ func (l *lab) Restart(ctx context.Context) error {
...
+	 	wg.Add(1)
+	 	go func (){
+	 		// closing VMs....
+	 		defer wg.Done()
+			if err := lab.vm.Close(); err!=nil{
+				log.Error().Msgf("Error on Close function in lab.go %s",err)
+			}
+		}()
+	 }
+	 go func() {
```
kdhageman: no `wg.Add(1)`?

lab/lab.go
```diff
@@ -208,24 +207,26 @@ func (l *lab) Restart(ctx context.Context) error {
...
-	}
-
-	if l.environment != nil {
-		wg.Add(1)
-		go closer(l.environment)
-	}
-
+	 for _, lab  := range l.frontends {
+	 	wg.Add(1)
+	 	go func (){
```
kdhageman: I would suggest to pass the lab as a parameter to the goroutine function something like this:
```
go func (lab lab.Lab){
	// do things with lab
}(lab)
```

mrturkmen06: What is the main reason of that ?

lab/lab.go
mrturkmen06: ups, added.

lab/lab.go
mrturkmen06: discussed.

Issue #246, opened 2019-08-26
white screen issue on unfinished events 


mrturkmen06:
Apparently,  [pull request] (https://github.com/aau-network-security/haaukins/pull/230) has a bug regarding to unfinished events on sec02, so if there are some existing files which are stored under events/ directory when an event is initialised, waiting screen fall into a status which does not assign VM to unfinished events. Therefore, a fix can provide us opportunity to run  new patch version 1.6.3 on sec02 for now until re-factorization process is done. When new release published, the version on sec02 can also updated. 
There is no issue discovered on newly created events.

mrturkmen06:
Regarding to commits which have been done to 1.6.3 branch, assigning lab to unfinished event issue has been resolved. However, lab.Hub will eventually change with refactor-lab-hub branch, since it is much clear and understandable.

PR #248, opened 2019-08-26
White scree minor fix

mrturkmen06:
@jbjjbjjbj  has been edited the code and it can resolve the issue #246 .


app/client/cli/event.go
```diff
@@ -157,8 +156,8 @@ func (c *Client) CmdEvents() *cobra.Command {
 			}
 
 			f := formatter{
-				header: []string{"EVENT TAG", "NAME", "# TEAM", "# EXERCISES", "CAPACITY"},
-				fields: []string{"Tag", "Name", "TeamCount", "ExerciseCount", "Capacity"},
+				header: []string{"EVENT TAG", "NAME", "# TEAM", "# EXERCISES", "CAPACITY", "CREATION TIME"},
```
tpanum: It is really messy when isolated features and bug fixes are handled in a non-version branch.

daemon/daemon.go
```diff
@@ -56,6 +56,7 @@ var (
 
 const (
 	mngtPort = ":5454"
+	dateTimeFormat = "2006-01-02 15:04:05"
```
tpanum:
```suggestion
	displayTimeFormat = "2006-01-02 15:04:05"
```

daemon/daemon.go
```diff
@@ -612,10 +612,9 @@ func (d *daemon) RestartTeamLab(req *pb.RestartTeamLabRequest, resp pb.Daemon_Re
 		return err
 	}
 
-	lab, err := ev.GetHub().GetLabByTag(req.LabTag)
-
-	if err != nil {
-		return err
+ 	lab,ok:=ev.GetLabByTeam(req.TeamId);
```
tpanum: Again, `go fmt` or `goimports` needs to be added to your IDE, so it formats the code when saving it.

daemon/daemon.go
```diff
@@ -700,13 +699,13 @@ func (d *daemon) ListEvents(ctx context.Context, req *pb.ListEventsRequest) (*pb
...
 	for _, event := range d.eventPool.GetAllEvents() {
 		conf := event.GetConfig()
-
 		events = append(events, &pb.ListEventsResponse_Events{
 			Name:          conf.Name,
 			Tag:           string(conf.Tag),
 			TeamCount:     int32(len(event.GetTeams())),
 			ExerciseCount: int32(len(conf.Lab.Exercises)),
 			Capacity:      int32(conf.Capacity),
+			CreationTime : conf.StartedAt.Format(dateTimeFormat),
```
tpanum:
```suggestion
			CreationTime : conf.StartedAt.Format(displayTimeFormat),
```

event/event.go
```diff
@@ -57,6 +57,7 @@ func (eh *eventHost) CreateEventFromEventFile(ctx context.Context, ef store.Even
 		return nil, err
 	}
 
+
```
tpanum: This line can be removed.

event/event.go
```diff
@@ -191,15 +192,18 @@ func (ev *event) Start(ctx context.Context) error {
 		log.Warn().Str("Team 0 ", ev.store.GetTeams()[0].Name).Msg("0 indexed team....")
 	}
 	for _, team := range ev.store.GetTeams() {
-		if err := ev.AssignLab(&team); err != nil {
-			//log.Error().Msg("Error while issuing error to team.. Check out assignlab function... ")
+
+		ev.teamQueue <- team
+		if err := ev.processQueue(); err != nil {
+			//log.Error().Err(err).Msg("Error while issuing error to team.. Check out assignlab function... ")
```
tpanum: Remove this comment.

svcs/guacamole/intercept.go
```diff
@@ -309,10 +211,10 @@ func (gtl *guacTokenLoginEndpoint) Intercept(next http.Handler) http.Handler {
 			}
 			u, err := gtl.users.GetUserForTeam(t.Id)
 			if err != nil {
-				log.Warn().
```
tpanum: Why is this commented out?

event/event.go
```diff
@@ -154,7 +155,7 @@ func NewEvent(ctx context.Context, ef store.EventFile, hub lab.Hub) (Event, erro
 		closers:       []io.Closer{ctf, guac, hub, keyLoggerPool},
 		dockerHost:    dockerHost,
 		keyLoggerPool: keyLoggerPool,
-		teamQueue:     make(chan *store.Team, conf.Capacity),
+		teamQueue:     make(chan store.Team, conf.Capacity),
```
tpanum: Is this safe? And why change it? Are we sure no pass-by-value problems will occur due to this change?

event/event.go
```diff
@@ -191,15 +192,18 @@ func (ev *event) Start(ctx context.Context) error {
 		log.Warn().Str("Team 0 ", ev.store.GetTeams()[0].Name).Msg("0 indexed team....")
 	}
 	for _, team := range ev.store.GetTeams() {
-		if err := ev.AssignLab(&team); err != nil {
-			//log.Error().Msg("Error while issuing error to team.. Check out assignlab function... ")
+
+		ev.teamQueue <- team
```
tpanum: Based on what `ev.processQueue`, why introduce a channel here? As I see it, nothing is concurrent and we are effectively inserting into a channel in order to instantly read from it?

Generally, `processQueue` seems like a strange work around of fixing issues with `lab.Hub`. Unless there is something I have severely missed, I would suggest adapting the direction of the `refactor-labhub` branch instead.

svcs/guacamole/intercept.go
```diff
@@ -14,216 +14,118 @@ import (
 )
 const waitingHTMLTemplate = `
```
tpanum: I think this direction is also conceptually not ideal. Assigning labs to existing events (on boot) should ideally just wait with giving users access to the frontend until every known team has assigned a lab. In the case of new events, people will be more-or-less instantly assigned a lab or should be notified that labs are either maxed out or unavailable during signup.

daemon/daemon.go
mrturkmen06: enabled filewatcher on golang. Thank you for the feedback :)

app/client/cli/event.go
mrturkmen06: Although I did not distinguish them properly, I agree.

svcs/guacamole/intercept.go
mrturkmen06: :( Actually, it is  commented out because waiting page has interval refresh rate of 10 seconds, after 10 seconds it refresh the page, but if there is no lab to assign the log will be printed out into daemon service log, until they assigned to a lab, same log statement created a lot of times, maybe 30 or 40 lines which increased redundancy on server side. Therefore, it is removed, but I accept that it is definitely right solution to do. Maybe having long refresh interval time or no auto refresh on waiting page might be more acceptable approach. However, again it is great to have feedbacks :)

event/event.go
jbjjbjjbj: Yes pointer would have been a lot better. I thought it was okay because the all the store.Team functions return store.Team as a value and not a pointer.

The reason i changes it was because then we loop `ev.store.GetTeams` and use a pointer, the underlining memory will change when the loop runs, and thus making all the queued teams the same.

I'm very welcome to feedback on how to solve this.

event/event.go
jbjjbjjbj: The channels are more used as a concurrently safe fifo queue, than to sync different tasks. 
Other structures could have achieved the same.

The ideas is that a team is added to the queue, and it calls the process queue.
If no teams where in the queue at that point it would just assign the team is just added, and then it's a bit stupid. 

But if there are no labs to assign it will exit early and that team will stay in the queue until the next time the `processQueue` is run.

The labhub will cal processQueue(through the hook) every time i creates a new lab thus assigning the new lab to a waiting team.

Yes the labhub design in refactor-labhub is much better, but i had the idea that this fix should be more short term.

svcs/guacamole/intercept.go
jbjjbjjbj: Yes this may be a better way than queuing the request, at least a lot simpler. 

However it should be handled a bit better in the frontend, as an error will undo all the information in the registration form. 
I imagine this is a lot easier then the amigo frontend is done.

event/event.go
tpanum: Ah, now it is more clear to me. The proccesQueue functions primarily as a signal for processor to perform some work while avoid blocking. The code is a little hard to grasp at first (just like the old messy version of `lab.Hub`). However, I still think the code ultimatively solve a problem with a poorly programmed `lab.Hub`. I think it is fine for this pull request. **But**, I think it would be valuable to add the refactored `lab.Hub`, as I believe it makes the code simplier (fewer lines of code), in addition to being easier to test.

mrturkmen06: We will eventually adapt refactor-labhub 👍🏻 since it is much clear as you mentioned.

PR #249, opened 2019-08-27
event restart command has been fixed

### LINKED ISSUE #156: Fix `event restart` command
Unable to restart resources of a team on an event, 
` hkn event restart <event-tag> <team-id>` command does not work as intended !

mrturkmen06:
#156  issue has been fixed and merged into white-scree-minor-fix branch in order to merge all changes into 1.6.0


daemon/daemon.go
```diff
@@ -612,10 +612,9 @@ func (d *daemon) RestartTeamLab(req *pb.RestartTeamLabRequest, resp pb.Daemon_Re
 		return err
 	}
 
-	lab, err := ev.GetHub().GetLabByTag(req.LabTag)
-
-	if err != nil {
-		return err
+ 	lab,ok:=ev.GetLabByTeam(req.TeamId);
```
tpanum: Please make sure that the code is either formatted by `go fmt` or `goimports`, most IDEs should do this on default as you save go files.

daemon/daemon_test.go
```diff
@@ -703,8 +703,6 @@ func TestListEvents(t *testing.T) {
 		count        int
 		err          string
 	}{
-		//{name: "Normal", count: 1},
```
tpanum: I think it is dangerous just to remove them. The big question is why are they commented out?

daemon/daemon.go
mrturkmen06: 👍🏻

daemon/daemon_test.go
mrturkmen06: The goroutines which are used in getServer  and grpc.DialContext creates problem somehow, I cannot say how but when I try to run the test gorutouni nill pointer memory error happened.

mrturkmen06:
The main change is that in  current 1.6.0 branch, lab is retrieved by [following command](https://github.com/aau-network-security/haaukins/blob/f95a8dcc0512fe51f6051496e827e4944a0d68c6/daemon/daemon.go#L615) 
``` 
lab, err := ev.GetHub().GetLabByTag(req.LabTag)

```
However, requested information from user was team id not lab tag, for this reason I have modified it into [this](https://github.com/aau-network-security/haaukins/blob/7d2fb1f60abaa3e3ecfe0d32f61a27867ebbf74e/daemon/daemon.go#L615)
```
lab,ok:=ev.GetLabByTeam(req.TeamId);
```

daemon/daemon_test.go
tpanum: Instead of removing tests, I believe we should try to understand why they fail. So diagnosing this is important from my perspective.

mrturkmen06: I will try to fix that 👍🏻 by synchronising the goroutines. :)

mrturkmen06: Apparently, getServer function which includes handler, creates problem when creationTime value added to listevents, I will try to find a solution to it.

mrturkmen06: fixed

PR #261, opened 2019-09-11
readme.md file updated

mrturkmen06:
Readme file updated any suggestion, addition, correction would be very welcome !


mrturkmen06:
Newly created README.md file can be seen in good format from here : https://github.com/aau-network-security/haaukins/tree/update-readme-file-%23260

mrturkmen06:
ok let's discuss it on monday.

Issue #211, opened 2019-07-08
Use go modules for managing versions of dependencies


kdhageman:

mrturkmen06:
Go modules integrated into Haaukins, however there is a problem with a package and actually it is declared on golang wiki page over [here](https://github.com/golang/go/wiki/Modules#how-can-i-resolve-parsing-gomod-unexpected-module-path-and-error-loading-module-requirements-errors-caused-by-a-mismatch-between-import-paths-vs-declared-module-identity)

mrturkmen06:
integrated to version [1.6.4](https://github.com/aau-network-security/haaukins/releases/tag/1.6.4)

Issue #264, opened 2019-09-17
enable description, hints and categories for challenges 


mrturkmen06:
In Haaukins platform, there is no description and hints for challenges, however as I noticed that those features can be enabled by adding hints, category and description information into YML configuration file of exercises. So eventually, the some existing features on CTFd official repo will opened in our side as well. There are some screenshots regarding to mentioned features. 

### Adding categories to exercises

![image](https://user-images.githubusercontent.com/13614433/65020162-e2e5a500-d92c-11e9-808c-bd57a9d32a73.png)

### Enabling hints  and description for exercises 

Example SS taken from https://demo.ctfd.io/ 

<img width="802" alt="Screenshot 2019-09-17 at 09 25 48" src="https://user-images.githubusercontent.com/13614433/65020337-4d96e080-d92d-11e9-8663-ca27870b229e.png">

mrturkmen06:
Enabled with [enable-description-#264](https://github.com/aau-network-security/haaukins/tree/enable-description-%23264) branch.

PR #277, opened 2019-09-28
Integrate go modules #211

mrturkmen06:


mrturkmen06:
Enabling go-modules into enable-description-#264 branch

PR #285, opened 2019-10-02
different formats of logo has been added

mrturkmen06:
- available formats of logo has been added to Haaukins platform.


mrturkmen06:
beta has been removed

PR #291, opened 2019-10-17
Fix register behaviour

mrturkmen06:
* quick fix to improper behaviour of register page
* releaser api key changed


svcs/ctfd/interception.go
```diff
@@ -677,3 +679,13 @@ func recordAndServe(next http.Handler, r *http.Request, w http.ResponseWriter, m
 
 	return rec.Result(), body
 }
+
+func isTeamExists(name,email string, ri *registerInterception) bool{
```
jbjjbjjbj: Why not use the GetTeamByEmail and GetTeamByName instead of looping through the teams.

mrturkmen06: Actually there is no particular reason and  yeah that would be much nicer, correct, will be changed. 👍🏻 Thanks for the comment

PR #304, opened 2019-10-25
remove default nic from vm

mrturkmen06:


svcs/ctfd/interception.go
```diff
@@ -305,11 +305,6 @@ func (ri *registerInterception) Intercept(next http.Handler) http.Handler {
 
 	updateRequest := func(r *http.Request, t *store.Team) error {
 		var err error
-		name := r.FormValue("name")
```
jbjjbjjbj: Why are these changes included?

mrturkmen06: It caused problems regarding to teams, "Unknown token" error started to appear, in addition for unfinished events, daemon could not assign all teams to events.

Issue #309, opened 2019-10-29
update-challenges-file-with-gRPC-call


mrturkmen06:
**Is your feature request related to a problem? Please describe.**

Yes, indeed ! When we have a new challenge where we should update exercises.yml file in order to include the new challenges for new events, we need to restart whole system from scratch which is really frustrating. 

__Proposal__

`hkn exercise update` will automatically reflect the changes on the file in background. 
This issue could be updated time-to-time, my idea is that when we deploy new challenges into sec01, the exercises file should also be updated automatically. In this situation, we can set a `crontab` to  update exercises file in daemon's memory. Or time to time we can run update command as proposed. 

**Describe the solution you'd like**

Having a command to update the exercises file or somehow integrate updates on exercises file with challenge deployment to sec01.

PR #310, opened 2019-10-30
Update event file

mrturkmen06:
#309


mrturkmen06:
The files are effected by gofmt command, so most of the changes we have in this requested caused by usage of gofmt.  Changes which should be reviewed are in following files. 
- app/client/cli/exercise.go 
- daemon/daemon.go
- event/event.go  
- store/exercise.go

store/exercise.go
```diff
@@ -283,11 +283,32 @@ type exercisestore struct {
...
 
+func (es *exercisestore) UpdateExercisesFile(path string) (ExerciseStore, error) {
+	var conf struct {
+		Exercises []Exercise `yaml:"exercises"`
+	}
+	f, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	err = yaml.Unmarshal(f, &conf)
```
jbjjbjjbj: Isn't this already implemented in https://github.com/aau-network-security/haaukins/blob/537393e9e1c1ee4f5a7efe5350b6cdd9b18e9ca9/store/exercise.go#L411.

I think it makes a lot more sense to have it here, maybe it can call into this function or something.

mrturkmen06: changed, thanks.

PR #314, opened 2019-10-31
error handling package added

mrturkmen06:


mrturkmen06:
JSON Logging will also be integrated under this branch

daemon/daemon.go
```diff
@@ -680,14 +693,17 @@ func (d *daemon) ResetExercise(req *pb.ResetExerciseRequest, stream pb.Daemon_Re
 	for _, t := range ev.GetTeams() {
 		lab, ok := ev.GetLabByTeam(t.Id)
 		if !ok {
-			stream.Send(&pb.ResetTeamStatus{TeamId: t.Id, Status: "?"})
+			if err:=stream.Send(&pb.ResetTeamStatus{TeamId: t.Id, Status: "?"}); err!=nil {
```
jbjjbjjbj: Hmm is it just me or did gofmt overlook this one.

errors/errors.go
```diff
@@ -0,0 +1,62 @@
...
+			msg = fmt.Sprintf("%v - %v:%v", msg, file, line)
+		}
+		e.Err = errors.New(msg)
+	}
+	for _, a := range args {
+		switch a := a.(type) {
+		case error:
+			e.Err = a
+		case string:
+			e.Err = errors.New(a)
```
jbjjbjjbj: When providing a err and string as in line 200, will it not just overwrite e.Err with the last one.

errors/errors.go
```diff
@@ -0,0 +1,62 @@
...
+		_, file, line, ok := runtime.Caller(1)
+		if ok {
+			msg = fmt.Sprintf("%v - %v:%v", msg, file, line)
+		}
+		e.Err = errors.New(msg)
+	}
+	for _, a := range args {
+		switch a := a.(type) {
+		case error:
+			e.Err = a
```
jbjjbjjbj: Should we set the Severity here?

errors/errors.go
```diff
@@ -0,0 +1,62 @@
...
+	"errors"
+	"fmt"
+	"github.com/sirupsen/logrus"
+	"runtime"
+)
+
+type Error struct {
+	function FCall
+	Err      error
+	Severity logrus.Level
```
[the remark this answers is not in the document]
mrturkmen06: True, zerolog will be used.

daemon/daemon.go
mrturkmen06: I think, gofmt did not applied :(

errors/errors.go
mrturkmen06: Actually, it will act according to type of given argument, so I do not see any overwrite situation here. For instance, it will be used as following... 
`daemon.go`

```
import ( 
   cError "github.com/aau-network-security/haaukins/errors"
)
func New(....) error {
const fCall cError.FCall = "daemon.New"

if err!= .... ;  err!=nil {
    return cError.E(fCall,err)
}
// or in case of we have defined constant error variables
if err!= .... ;  err!=nil {
    return cError.E(fCall,UnknownMethodErr)
}
}
```
I hope, I understood your statement.

errors/errors.go
mrturkmen06: In same cases, it could be handy to have loglevel as severity, but yeah, there should not be another logging package, either logzero or logrus should be chosen. I will stick with logzero.

errors/errors.go
jbjjbjjbj: I don't think i understand it correctly then. What happens if one provided a string and err like this

```
cError.E(op, err, fmt.Sprintf("unable to read users file: %s", conf.UsersFile))
```

Wouldn't the string just replace the err?

jbjjbjjbj: I don't think i understand it correctly then. What happens if one provided a string and err like this

```
cError.E(op, err, fmt.Sprintf("unable to read users file: %s", conf.UsersFile))
```

Wouldn't the string just replace the err?

mrturkmen06: Yeah, true in that case it will overwrite, but can't it be written as following ?  ; 
```
cError.E(op,fmt.Sprintf("unable to read users file: %s err: %s", conf.UsersFile,err))
```

jbjjbjjbj: Yes but why not have it function like https://godoc.org/github.com/pkg/errors#Wrapf.
So if you apply a string and err it will look something like this.

```
Applied string: applied error
```

mrturkmen06: so, do you think that providing Fcall into E is not useful at all ? or we do not need such a thing and just use wrapf ?

jbjjbjjbj: No the fcall is a good idea. But I think we should make it so that it accepts both err and strings and then combines them like wrapf

daemon/daemon.go
```diff
@@ -923,29 +945,33 @@ func (d *daemon) grpcOpts() ([]grpc.ServerOption, error) {
...
 				log.Warn().Msgf("Serving error: %s", err)
+				// todo: handle error
 			}
 
 			return
 		}
 
 		if err := http.ListenAndServe(fmt.Sprintf(":%d", d.conf.Port.InSecure), d.eventPool); err != nil {
-			log.Warn().Msgf("Serving error: %s", err)
+			fmt.Errorf("Listen and serve error %d ", err)
```
jbjjbjjbj: Why replace this log?

Btw. i dont think errorf prints anything.

errors/errors.go
```diff
@@ -0,0 +1,66 @@
...
+			e.Msg = a
+		case zerolog.Level:
+			e.Severity = a
+		}
+	}
+	if e.Err == nil && e.Msg != "" {
+		return errors.New(e.Msg)
+	}
+	if e.Err != nil && e.Msg != "" {
+		return errors.New(e.Error() + " " + e.Msg)
```
jbjjbjjbj: Can this be done with Errorf?

errors/errors_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func TestSeverity(t *testing.T) {
+	const functioncall FCall = "TestSeverity"
+	message := "error to test out severity of zerolog levels"
+	tests := []struct {
+		functionCall FCall
+		message      string
+		level        zerolog.Level
+	}{
+		{functionCall: functioncall, message: message, level: zerolog.ErrorLevel},
```
jbjjbjjbj: We should also test the other Levels i think

daemon/daemon.go
mrturkmen06: this is left from another experiment, mistakenly committed change , it is log.

errors/errors_test.go
mrturkmen06: Since E returns error type, it is not quite obvious how to test it. I think, previous version of E function was good enough (with logzero instead of logrus), since it is just a helper function and do not need to wrap all arguments that user typed.

jbjjbjjbj: But why set the log level in E then?

PR #331, opened 2019-11-27
Temp https connection

mrturkmen06:


mrturkmen06:
only one certificate could be used for both domains that we are dealing with, so one of cert field could be removed from config structure in daemon.go

PR #334, opened 2019-12-10
Modified Proto for Web Client

gianmarcomennecozzi:
I have modified the proto file in order to get the exercise tags on the ListEventsResponse message, instead of retrieve the count of them.


app/client/cli/event.go
```diff
@@ -155,7 +155,7 @@ func (c *Client) CmdEvents() *cobra.Command {
 
 			f := formatter{
 				header: []string{"EVENT TAG", "NAME", "# TEAM", "# EXERCISES", "CAPACITY", "CREATION TIME"},
```
jbjjbjjbj: We should remove the # before EXERCISES because it's not a count anymore.

gianmarcomennecozzi: Yeah you are right

mrturkmen06:
Another pull request will be created from 1.6.8 into x-web-client. This is not true one.

PR #346, opened 2020-01-25
Implemented virtual.Suspend

jbjjbjjbj:
This adds a Suspend function and state to the virtual package.

Implementation and tests has been added to vbox and docker.

This can be used to implement time-to-live for labs.


virtual/docker/docker_test.go
```diff
@@ -37,8 +37,9 @@ func testCleanup(t *testing.T, c hkndocker.Container) func() {
 }
 
 func TestContainerBase(t *testing.T) {
-	// testing create
+	// testing create. Do a long sleep to keep it alive
 	c1 := hkndocker.NewContainer(hkndocker.ContainerConfig{
+		Cmd:   []string{"sleep", "1d"},
 		Image: "alpine",
 	})
```
mrturkmen06: Having a long sleep in CI might cause problems, I think, for the sake of tests, it could be skipped during CI test or sleep amount could be much less than 1 day.

virtual/vbox/vbox.go
```diff
@@ -146,6 +146,23 @@ func (vm *vm) Stop() error {
...
+			Msgf("Failed to suspend VM: %s", err)
+		return err
+	}
+
+	log.Debug().
+		Str("ID", vm.id).
+		Msgf("Suspended vm")
+
+	return nil
+}
```
mrturkmen06: There is no problem on saving the state of VM however have you check that it becomes available when you try to wake it up ?

mrturkmen06: On host, rather than test cases, it  acts differently.

virtual/vbox/vbox_test.go
```diff
@@ -53,20 +54,44 @@ func TestVmBase(t *testing.T) {
...
+		t.Fatalf("expected virtual machine to have been stopped")
 	}
 
+	// Check state using info command
+	if info := vm.Info(); info.State != virtual.Suspended {
+		t.Fatalf("expected virtual machine to be suspended")
+	}
+
+	teststart()
+
```
mrturkmen06: Probably, this will gonna work, however  when there is a interaction between vboxmanager and OS on waking up an machine which is  in saved state, causes problems, it is unable to wake the machine up correctly. It is something related to vboxmanage and OS, not from your code.

virtual/docker/docker_test.go
jbjjbjjbj: Sorry for the late comment. 

It is not meant the run for one day, pretty sure stopping is part of the test.

mrturkmen06: Ah, ok then. I thought it is one day.

virtual/vbox/vbox.go
jbjjbjjbj: Pretty sure i tested i on my own computer, and that it worked fine.
But will test it again just to be sure.

mrturkmen06: I can provide you an access to test environment, if you would like to test it on server instead of your local, send me your public key.

virtual/vbox/vbox_test.go
jbjjbjjbj: Hmm didn't experience that.
Will try it again to be sure i didn't miss something.

jbjjbjjbj:
Hey sorry for my late response, got lost in my github notification.

Thanks for the feedback, will test it again.
Should also test it on other computers than my laptop.

mrturkmen06:
Works like a charm ! :)

PR #356, opened 2020-02-24
Add privilege to users

mrturkmen06:


mrturkmen06:
In this pull request, role based permission added, admin > user > non-priv-user , clarified with permissions and capabilities.

PR #357, opened 2020-02-26
first version of scheduled events 

mrturkmen06:
- First  draft of starting events has been added. 
- Requires different test cases using sec01
- Error on getunfinishedevents : needs to be fixed.


daemon/proto/daemon.proto
```diff
@@ -64,9 +65,15 @@ message CreateEventRequest {
   repeated string exercises = 4;
   int32 available = 5;
   int32 capacity = 6;
-  string finishTime = 7;
+  string startTime = 7;
+  string finishTime = 8;
+  bool isBooked = 9;
```
gianmarcomennecozzi: I am not sure where `isBooked` is used or set. I am not using it in the WebClient and seems it is not used in the Client either. in this case it can be deleted.

mrturkmen06: true, it can be deleted.

mrturkmen06: I was thinking to make something else at that point that's why I added. I guess :)

mrturkmen06:
I am thinking to make more tests on this branch afterwards merge it.

mrturkmen06:
For event list command, booked events are also added.

PR #359, opened 2020-03-08
Amigo

gianmarcomennecozzi:
Amigo structure re-built:
- new CSS
- NPM version and packages updaged
- Removed all useless file


store/myevent.go
```diff
@@ -0,0 +1,68 @@
+package store
```
mrturkmen06, line 1: I assume that this file is created for test purposes ?  If it is not written for test purposes why it is called `myevent.go`

store/team.go
```diff
@@ -103,22 +90,36 @@ func (es *teamstore) GetTeamByToken(token string) (*haaukins.Team, error) {
...
+
 	es.emails[email] = t.ID()
 	es.teams[t.ID()] = t
 	es.m.Unlock()
 
 	return nil
 }
 
+//todo does it create a token or it save it in teamstore? Ahmet. I thinnk it saves it. so i can change hte name of this function
 func(es *teamstore) CreateTokenForTeam (token string, in *haaukins.Team) error {
```
mrturkmen06, line 123: Yeah, the name of the function does not fully represent what it is doing. It just saves the existing given token to teamstore.

store/myevent.go
gianmarcomennecozzi: Yeah it is :)

mrturkmen06:
`the_real_amigo` branch migrated to `develop-a`, we agreed that this merge will not gonna happen.

PR #366, opened 2020-04-10
Keylogger

gianmarcomennecozzi:
Reproduced keylogger after haaukins refactoring


gianmarcomennecozzi:
I expected the CI failing for the tests. I will clean up the dependencies anyway.

PR #378, opened 2020-04-20
gh_actions test added

mrturkmen06:
- github actions test integration added

- Since the working principle of travis and github actions differ between each other, the test called `TestBasicEnvironment` passes the test, however, it sometimes fails and sometimes passes on github actions, that's why I have added ignore keyword only for github actions, so the test needs to be passed on travis CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: Test
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


mrturkmen06:
Note that, seems the code_coverage_token does not work very well, I will keep fixing it, it does not effect test cases however nice to have integrated code coverage report.

mrturkmen06:
I have replaced code coverage token in secrets, now it should be fine, there should not be any complain regarding to codecov report.

PR #382, opened 2020-04-26
Adjust Style

gianmarcomennecozzi:
Fixed css style errors seens in https://alpha.dev.ntp-event.dk/


gianmarcomennecozzi:
Before to deploy it to sec01, we should change the points given to each challenge.
Since the new scoreboard table show the points, my suggestion is to have a scale from 0 - 100 in which every challenge has a different score.

mrturkmen06:
> Before to deploy it to sec01, we should change the points given to each challenge.
> Since the new scoreboard table show the points, my suggestion is to have a scale from 0 - 100 in which every challenge has a different score.
> @mrturkmen06

Then, I think, it could be logical to multiply each challenge with `x5` or `x4`, let's say if we have a challenge of 20 points max, then it could be logical to multiple each challenge with `5` and same for others as well.

gianmarcomennecozzi:
yeah it makes sense. I think it looks nicer if we have all the challenges with a different score

svcs/amigo/table.go
```diff
@@ -10,6 +11,15 @@ type Message struct {
 	Values  interface{} `json:"values"`
 }
```
mrturkmen06, line 13:
```suggestion
var (
       ChallengeCategories = [5]string{"Web exploitation", "Forensics", "Cryptography", "Binary", "Reverse Engineering"}
)
```

mrturkmen06:
> yeah it makes sense. I think it looks nicer if we have all the challenges with a different score

OK, it means there is no need to change anything in challenges file on server, nice.

svcs/amigo/table.go
```diff
@@ -21,20 +31,36 @@ type TeamRow struct {
...
 	TeamRow []TeamRow	`json:"teams"`
 }
 
 func (fd *FrontendData) initTeams(teamId string) []byte {
 
 	teams := fd.ts.GetTeams()
 	rows := make([]TeamRow, len(teams))
+	challenges := make([]ChalRow, 5)
+
+	category := [5]string{"Web exploitation", "Forensics", "Cryptography", "Binary", "Reverse Engineering"}
```
mrturkmen06, line 44: It could be nice to have this category string as variable at the beginning of the file. I have added the code above.

svcs/amigo/table.go
```diff
@@ -21,20 +31,36 @@ type TeamRow struct {
...
+	for i, c := range category{
```
mrturkmen06, line 45:
```suggestion
	
	for i, c := range ChallengeCategories {
```

mrturkmen06: It will be modified automatically if you directly accept it.

svcs/amigo/table.go
mrturkmen06: It will be modified automatically if you directly accept it.

svcs/amigo/table.go
```diff
@@ -21,20 +31,36 @@ type TeamRow struct {
...
+	}
 
 	chalsHelper := make([]store.FlagConfig, len(fd.challenges))
-	chals := make([]string, len(fd.challenges))
 	for j, c := range fd.challenges {
 		chalsHelper[j] = c
-		chals[j] = c.Name
+		for i, rc := range challenges{
+			if rc.Category == c.Category{
+				fmt.Print(c.Name)
```
mrturkmen06, line 57:
```suggestion
				
```

svcs/amigo/table.go
```diff
@@ -49,30 +75,13 @@ func (fd *FrontendData) initTeams(teamId string) []byte {
...
-	//msg := Message{
-	//	Message: "teams",
-	//	Values:  rows,
-	//}
-	//rawMsg, _ := json.Marshal(msg)
-	//
-	//return rawMsg
 }
 
 func TeamRowFromTeam(t *store.Team, chals []store.FlagConfig) TeamRow {
```
mrturkmen06, line 87: I think, the name of the function is a little bit odd. `TeamInfo` could be much nicer.

PR #385, opened 2020-04-28
Hotfix/split into parts

mrturkmen06:
- Some files restructured 
- Some functions refactored or removed. 
- Package based movements is done.
- Duplicate codes removed (`rawevent`)

I think, there are still things to change, especially, refactoring on `store` module would be nice for second step. 

I am proposing changes right now to `develop-a` branch, then I will try to keep changing some stuffs . But for now, it can be merged.


mrturkmen06:
I suggest you to look at the changes commit by commit, instead of reviewing them all in one. @gianmarcomennecozzi

Issue #386, opened 2020-04-28
Amigo Frontend Bugs


gianmarcomennecozzi:
- [x] Sort challenges per points in the scoreboard page

- [x] sort teams in the team page

- [x] make fix width columns scoreboard table

- [ ] insert loading logo while the teams and scoreboard page are fetching data from the server (Not sure, let check first how fast it is with the new changes)

- [x] change challenges points in order to have all of them different

PR #387, opened 2020-04-29
Fix frontends bugs #386

gianmarcomennecozzi:
i made changes in the scoreboard page and i delete some JS code in order to make the loading page faster


svcs/amigo/resources/private/js/src/components/Scoreboard.vue
```diff
@@ -1,15 +1,17 @@
 <template>
     <div class="table-responsive">
-        <table class="table table-striped ">
+        <table class="table table-striped" id="scoreboardtable" style="min-width: 1500px!important;">
```
mrturkmen06, line 3: why it has min-width of 1500px ?

svcs/amigo/resources/private/js/src/components/Scoreboard.vue
```diff
@@ -47,40 +48,31 @@
...
+                    case "Web exploitation":
+                        return "Web E."
+                    case "Forensics":
+                        return "For.."
+                    case "Cryptography":
+                        return "Cry.."
+                    case "Binary":
+                        return "Bin.."
+                    case "Reverse Engineering":
+                        return "R. Eng."
```
mrturkmen06, line 74: I think, it is quite nice to have shorter texts in scoreboard however, it is also irritating when you do not know  the exact meaning of `For...`  , it could be much nicer when you hover on `For...` it should show `Forensics` and etc. As you have already implemented for challenge names. 

Do not you think it is required ?

svcs/amigo/table.go
```diff
@@ -33,40 +36,47 @@ type TeamRow struct {
...
 }
 
 func (fd *FrontendData) initTeams(teamId string) []byte {
 
 	teams := fd.ts.GetTeams()
 	rows := make([]TeamRow, len(teams))
-	challenges := make([]ChalRow, 5)
-
+	var challenges []ChalCategory
+	var realChallenges []ChalCategory
```
mrturkmen06, line 48: why we have same things with different names ?

svcs/amigo/table.go
```diff
@@ -33,40 +36,47 @@ type TeamRow struct {
 }
 
 type Scoreboard struct {
-	Chals 	[]ChalRow	`json:"challenges"`
+	Chals 	[]ChalCategory	`json:"challenges"`
```
mrturkmen06, line 39: Naming of variables is a little  bit  confusing. Do  not you think that having following make much sense ?

```suggestion
	ChalCats 	[]ChalCategory	`json:"chal-categories"`
```

svcs/amigo/resources/private/js/src/components/Scoreboard.vue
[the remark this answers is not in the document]
gianmarcomennecozzi: It is used to make the table responsive for different screens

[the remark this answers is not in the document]
mrturkmen06: Ok, but why 1500px not 1800px for instance ? I would like to understand it.

svcs/amigo/resources/private/js/src/components/Scoreboard.vue
gianmarcomennecozzi: hm.. it could be nice. i will see what i can do

svcs/amigo/table.go
```diff
@@ -108,20 +120,6 @@ func TeamRowFromTeam(t *store.Team, chals []store.FlagConfig) TeamRow {
...
-//	}
-//
-//	return TeamRow{
-//		Id:          t.ID(),
-//		Name:        t.Name(),
-//		Completions: completions,
-//	}
-//}
-
 type ChallengeF struct {
```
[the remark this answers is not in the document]
gianmarcomennecozzi: hmm not sure. it's just a useful `struct` used to send data to the frontend.

svcs/amigo/table.go
gianmarcomennecozzi: That's quite important. basically i remove the categories that have not any challenges inside.

gianmarcomennecozzi: creating a new variable was the only way to do that

svcs/amigo/table.go
gianmarcomennecozzi: Yeah i agree. I am going to make some comments on them

svcs/amigo/table.go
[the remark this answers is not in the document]
mrturkmen06: but when I read it, I do not give any meaning to it. It is not a good naming unfortunately.

svcs/amigo/resources/private/js/src/components/Scoreboard.vue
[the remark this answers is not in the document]
gianmarcomennecozzi: hmmm, there is not much to understand. I have tried different sizes and seems that 1500px is good. higher is that number and less responsive will be the table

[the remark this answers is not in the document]
mrturkmen06: oh, ok. I wonder that there could be automatic way of it, but anyway not very important, I just wanted to see how it is clarified.

svcs/amigo/table.go
mrturkmen06: but do we have categories which has no challenges ?

PR #389, opened 2020-05-02
API direct to guacamole from Learning Material

gianmarcomennecozzi:
in order to create a link in between learning materials and challenges, an API has been developed. 
A user can make up to 4 requests per hour (this functionality has to be implemented).
we track the user requests by his ip and every requests has made by each user has been mapped to the challenge tag.
at every request a loading page is shown and refreshing the page will not start another process in background (i have set a cookie while the loading page is running), when the process in background finish, the user will be automatically redirected to guacamole/
if the user makes another request with the same challenges, he will be redirected immediately to his guacamole created without creating another environment.


api/api.go
```diff
@@ -0,0 +1,200 @@
...
+			w.Write([]byte(waitingHTMLTemplate))
+			return
+		}
+
+		port, _ := clientR.ports[challengesFromLink]
+
+		authC := http.Cookie{Name: "GUAC_AUTH", Value: cookie , Path: "/guacamole/"}
+		http.SetCookie(w, &authC)
+		http.SetCookie(w, &http.Cookie{Name: "session", MaxAge: -1}) //Remove the cookie
+		host := fmt.Sprintf("http://localhost:%s/guacamole", strconv.Itoa(int(port)))
```
kdhageman, line 120: This cannot work in production right? Redirecting to localhost

api/api.go
```diff
@@ -0,0 +1,200 @@
...
+		//Bad request (challenge tags don't exist)
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Header().Set("Content-Type", "text/html; charset=utf-8")
+			w.Write([]byte(errorHTMLTemplate))
+			return
+		}
+
+		//check if the user exists
+		clientR, ok := lm.crs.clientsR[r.Host]
```
kdhageman, line 81: What if an event is hosted on a high school that has a NAT so that all traffic comes from the same IP address; then only a single environment exists. As I suggested before, use cookies for this.

Also, rate limiting based on IP address suffers from the same problem, so this functionality simply will not work in Haaukins

api/client.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	crs.clientsR[host] = cl
+	return cl
+}
+
+type ClientRequest struct {
+	username 		string
+	password 		string
+	cookies			map[string]string				//map cookie with challenges tag
```
kdhageman, line 38: Theoretically speaking, only a single cookie should be required; each user gets a single cookie, then the state of this user is associated with the cookie. There is a natural distinction across events anyway I think, because (if I recall correctly) cookies apply to a full domain (i.e. `test.ntp-event.dk`), so cookies are not shared across events.

Then again, I don't know the exact implementation details of Haaukins, so this may require some refactoring to get it working right.

api/environment.go
```diff
@@ -0,0 +1,148 @@
...
+
+type environment struct {
+	timer		*time.Time
+	challenges 	[]store.Tag
+	lab 		lab.Lab
+	guacamole 	guacamole.Guacamole
+	guacPort 	uint
+}
+
+func (e environment) Assign(client *ClientRequest) error {
```
kdhageman, line 25: This seems like a re-use of existing code, right? 
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

daemon/daemon.go
```diff
@@ -389,5 +390,10 @@ func (d *daemon) Run() error {
 	reflection.Register(s)
 	log.Info().Msg("Reflection Registration is called.... ")
 
+	learningMaterialAPI, err := api.NewLearningMaterialAPI()
+	if err == nil {
+		go http.ListenAndServe(":8888", learningMaterialAPI.Handler())
```
kdhageman, line 395: Can't you add this handler to the existing Haaukins Mux that also serves all events. Why another web server?

Issue #396, opened 2020-05-07
unable to run haaukins due to handling certs


mrturkmen06:
**Describe the bug**
Seems after updates to haaukins-store, haaukins is also required to have some updates regarding to creating database client from Haaukins side. 
Handling of certificates in Haaukins prevents to communicate with haaukins-store when https used for development purposes. 

**To Reproduce**
Enable HTTPS in both haaukins and haaukins-store, then try to connect db, you will encourage with the error. 

**Expected behavior**
There should not any issue or error in between store and haaukins

PR #397, opened 2020-05-07
Hotfix/fix to dev certs #396

mrturkmen06:


mrturkmen06:
> It's good to have a `config` file so we just need to change it in order to let haukins run in different machines but at the same time i think that all those certificates might be a problem when some one what to run haaukins in his local machine.
> With this pull request we have basically added more complexity to run haaukins in local.
> 
> N.B the file `svcs/amigo/app/main.go` have to changed a little bit in order to have the possibility to get data from the DB in `sec01`

Yeah,  but as I mentioned you over slack and I noted on the code, handling of configurations and certs will be altered soon (I hope).

PR #399, opened 2020-05-09
listevents bug fixed & certs handling for cli changed a little bit

mrturkmen06:
- local certificates fetched directly from net (for only development purposes), if not exists 
- minor bug regarding to listevents fixed, daemon do not crash anymore


daemon/event.go
```diff
@@ -189,7 +189,9 @@ func (d *daemon) ListEvents(ctx context.Context, req *pb.ListEventsRequest) (*pb
 			Exercises: 	   strings.Join(exercises, ","),
 			Capacity:      int32(conf.Capacity),
 			CreationTime:  conf.StartedAt.Format(displayTimeFormat),
-			FinishTime:    conf.FinishExpected.Format(displayTimeFormat),
+			// There is no finishexpected field for develop-a branch.
+			// When booking is applied on this branch it can be added as well.
```
gianmarcomennecozzi, line 193: That was the problem

mrturkmen06: The list events function was not able to get FinishExpected field, (apparently there is no such a field for the events which are saved to db ). We were using finishexpected field for the events which are booked.

mrturkmen06: When events are listed from command line, the list events was checking 

> Certificates issues are always a problem for us. Hopefully we are in the good way in order to don't deal with them for a while. It's good you are taking care of this problems.

Let me clarify some points regarding to certificates, Haaukins was not supporting to have HTTPS connection for local development and the previous certificate issues are caused due to renewing existing certificates. Now, we are trying to have HTTPS connection both in prod and dev environment that's why it created a new issue.

PR #402, opened 2020-05-20
Hotfix/troubleshooting tmp err

mrturkmen06:
- spaces trimmed for registration and login forms on amigo 91f3b73
- duplicate getevents deleted 52d8786
-  check added here 795e83f
- troubleshooting doc introduced with initial information bb44314


mrturkmen06:
> I have tested this branch in local and seem that `strings.TrimSpace` dosen't work as expected. The problem of white space is still here. You should find another way to solve it i guess

Instead it is working, `TrimSpace` does not mean that it will remove spaces between words, it will gonna remove the spaces before and after the word.  Here you go; 

https://play.golang.org/p/A5Z75hurO_4

@gianmarcomennecozzi 

The problem was with that particular topic is that for instance I have an account called "mrturkmen", if I would like to register with "mrturkmen ", I was able to register, however, after TrimSpace, it is not possible. That was the our problem regarding to team names.   

However, I realized that when we are saving team into teamstore, we do not check usernames, we only rely on `emails` that's why it is possible to create multiple teams with same name. I think, the problem is more than TrimSpace.

gianmarcomennecozzi:
Okay then i think its good to remove spaces also from the inserted uername and check if that username has been already choosen.

mrturkmen06:
> Okay then i think its good to remove spaces also from the inserted uername and check if that username has been already choosen.

I have already added it, check [this line] (https://github.com/aau-network-security/haaukins/blob/21656b82cb5978379329108e003082b07b40b48d/store/team.go#L88).

PR #403, opened 2020-05-21
more funcs to user management added

mrturkmen06:
These changes are made for interacting and managing users who are using Haaukins to create and manage events. 

- Delete users 
- Change user password
- List existing users 


daemon/proto/daemon.proto
```diff
@@ -29,6 +33,22 @@ message Team {
   string Id = 1;
 }
 
+message DestroyUserRequest {
+  string username = 1;
+}
+
+message DestroyUserResponse {
+  string message = 1;
```
gianmarcomennecozzi, line 41: Add error parameter as well. we have to handle the response on the webclient.

daemon/proto/daemon.proto
```diff
@@ -29,6 +33,22 @@ message Team {
...
+}
+
+message UpdatePasswdRequest{
+  string username = 1;
+  string password = 2;
+}
+
+message UpdatePasswdResponse {
+  string message = 1;
```
gianmarcomennecozzi, line 50: Add error parameter as well here as well. we have to handle the response on the webclient.

daemon/user.go
```diff
@@ -84,3 +91,98 @@ func (d *daemon) InviteUser(ctx context.Context, req *pb.InviteUserRequest) (*pb
...
+func (d *daemon) DestroyUser(ctx context.Context, request *pb.DestroyUserRequest) (*pb.DestroyUserResponse, error) {
+
+	requester, err := getUserFromIncomingContext(ctx)
+	if err != nil {
+		log.Warn().Msgf("User credentials not found ! %v  ", err)
+		return &pb.DestroyUserResponse{}, err
+	}
+
+	if !requester.SuperUser {
+		return &pb.DestroyUserResponse{}, NoPrivilegeToDelete
```
gianmarcomennecozzi, line 137: Modify the response in case of error according the proto file changed (for each error response)

daemon/proto/daemon.proto
mrturkmen06: Message is used for both success or error  which ever comes first.

daemon/proto/daemon.proto
mrturkmen06: Message is used for both success or error which ever comes first.

mrturkmen06:
@gianmarcomennecozzi 
It works !  Please rename your `users` file with `users.yml.bckp`, and launch haaukins daemon,  then you can see that it is working, functions are passing from tests as well. The reason of not working in your case could be that, the `users.yml` file is outdated, I mean, we have more variable then just `username, password`.  Check [here](https://github.com/aau-network-security/haaukins/blob/7bfebccfc6f049853b1dfc2d6e906d492d536858/store/user.go#L30)

daemon/proto/daemon.proto
gianmarcomennecozzi: it's better to have error message anyway. i think easier to manage in the webclient.

mrturkmen06: @gianmarcomennecozzi, I am having difficulties to understand apparently, could you briefly explain me what would be the advantage of it ? How it will differ in webclient side to have this one ; 
```
message DestroyUserResponse {
  string message = 1; 
  string err = 2; 
}
```
instead 

```
message DestroyUserResponse {
  string message = 1;
}
```
or does it break some rules that we defined earlier and I do no remember now ?

gianmarcomennecozzi: the advantage of it is on the webclient. when i parse the response on the webclient how can i understand if the response is ok or not ?
if we have `message` and `err` parameters its easy to check whatever we got an error from haaukins or not.

gianmarcomennecozzi: thats why in `login` and `signup` response messages we have that `error` parameter in the proto file

mrturkmen06: Ok, normally, I do not prefer to have it, however, I do not want to keep you annoyed :) as well. However, I guess, only difference will be just checking message instead of error, 

Example

```
 daemonclient.loginUser(getRequest, {}, (err, response) => {
         if (err == null && response.getError() == '') {
                      localStorage.setItem('user', response.getToken());
                      localStorage.setItem('user-email', username);
                      router.push(this.returnUrl)
          }else{
               this.error = err || response.getError();
               this.submitted = false;
               this.loading = false;
            }
    });

```
into this, only difference will be ` this.message = err || response.getMessage();`

```

 daemonclient.loginUser(getRequest, {}, (err, response) => {
         if (err == null && response.getError() == '') {
                      localStorage.setItem('user', response.getToken());
                      localStorage.setItem('user-email', username);
                      router.push(this.returnUrl)
          }else{
               this.message = err || response.getMessage();
               this.submitted = false;
               this.loading = false;
            }
    });

```

Ok, I am going to change it into error, I do not want to create a barrier in front of you @gianmarcomennecozzi  :)

gianmarcomennecozzi: ohh wait. i miss some pieces of code. i have seen u were returning error as well. than u dont have to change it

mrturkmen06: Yeah, that's what I was trying to express actually @gianmarcomennecozzi.

mrturkmen06: @gianmarcomennecozzi, but keep in mind that you need to check content of the error message, because I am sending error as message and `success` as message in some cases. Like here: https://github.com/aau-network-security/haaukins/blob/7bfebccfc6f049853b1dfc2d6e906d492d536858/daemon/user.go#L176

gianmarcomennecozzi:
You are right.. seems they are working now :)

PR #409, opened 2020-05-25
Feature/integrate sleep mode

### LINKED ISSUE #416: extend sleep functionality to event base
Currently we are having sleep functionality team based, however there might be cases which we might need to have sleep functionality event based. These cases could be summarized as following; 

times to have events which we do not have enough resource to run (for bigger events) then it means that we should be able to suspend one of event in order to increase usability of the resources we have. 

restoring event function is handy to do not lose any team member or data regarding to that event. 
-

mrturkmen06:
Close #416


daemon/daemon.go
```diff
@@ -346,6 +345,22 @@ func (d *daemon) grpcOpts() ([]grpc.ServerOption, error) {
 	return []grpc.ServerOption{}, nil
 }
 
+func (d *daemon) Sleep() error {
+	ticker := time.NewTicker(5 * time.Hour) // every 5 hour check teams on each event
```
gianmarcomennecozzi, line 349: Either pass this value on the config file or document it somewhere if we agree to leave it here as static value. We are going  to forget about it if left.

daemon/proto/daemon.proto
```diff
@@ -7,7 +7,7 @@ service Daemon {
   rpc ListUsers (Empty) returns (ListUsersResponse) {}
   rpc ChangeUserPasswd (UpdatePasswdRequest) returns (UpdatePasswdResponse) {}
   rpc DestroyUser (DestroyUserRequest) returns (DestroyUserResponse) {}
-
+  rpc SetTeamSuspend (SetTeamSuspendRequest) returns (Empty) {}
```
gianmarcomennecozzi, line 10: is this one used to suspend and resume at the same time ?

daemon/team.go
```diff
@@ -2,11 +2,20 @@ package daemon
...
+
+	"github.com/aau-network-security/haaukins/virtual"
+
 	pb "github.com/aau-network-security/haaukins/daemon/proto"
 	"github.com/aau-network-security/haaukins/store"
 	"github.com/rs/zerolog/log"
 )
 
+const (
+	INACTIVITY_DURATION = 8 // in hours
```
gianmarcomennecozzi, line 16: lets document it somewhere or pass this value from config file.

daemon/team.go
```diff
@@ -64,4 +98,40 @@ func (d *daemon) RestartTeamLab(req *pb.RestartTeamLabRequest, resp pb.Daemon_Re
 	}
 
 	return nil
-}
\ No newline at end of file
+}
+
+// suspendTeams function will check all teams in all events
+// then suspend their resources if lastAccessTime is higher than INACTIVITY_DURATION
+func (d *daemon) suspendTeams() error {
```
gianmarcomennecozzi, line 105: cool :)

daemon/team.go
```diff
@@ -64,4 +98,40 @@ func (d *daemon) RestartTeamLab(req *pb.RestartTeamLabRequest, resp pb.Daemon_Re
...
+			if !t.LastAccessTime().IsZero() {
+				difference := math.Round(now.Sub(t.LastAccessTime()).Minutes()) // get in rounded format in hours
+				if difference > INACTIVITY_DURATION {
+					lab, ok := e.GetLabByTeam(t.ID())
+					if !ok {
+						return UnknownTeamErr
+					}
+					// check if it is already suspended or not.
+					lab.InstanceInfo()
+					for _, instanceInfo := range lab.InstanceInfo() {
```
gianmarcomennecozzi, line 122: why calling `lab.InstanceInfo()` twice ?

virtual/docker/docker.go
```diff
@@ -351,8 +351,31 @@ func (c *container) Start(ctx context.Context) error {
...
+
+	case virtual.Suspended:
+		fmt.Println("suspended")
+	case virtual.Error:
+		fmt.Println("error")
+	default:
+		// freebsd, openbsd,
+		// plan9, windows...
+		fmt.Println("no state retrieved")
+	}
```
gianmarcomennecozzi, line 369: remove all this prints

svcs/guacamole/keylogger_test.go
```diff
@@ -6,16 +6,20 @@ package guacamole_test
...
 	"testing"
 	"time"
+
+	"github.com/aau-network-security/haaukins/store"
+
+	"github.com/aau-network-security/haaukins/svcs/guacamole"
 )
 
 func TestKeyLogger(t *testing.T) {
+	// todo: could not understand why failed needs to be fixed, not urgent
```
gianmarcomennecozzi, line 21: i see, i think we can open an issue for that

daemon/daemon.go
mrturkmen06: yeah, exactly.

daemon/team.go
mrturkmen06: yup.

daemon/team.go
mrturkmen06: hmm, I really did not notice it, probably it is something forgotten from development process.

daemon/proto/daemon.proto
mrturkmen06: Yeah, in suspend request it checks whether [suspend value](https://github.com/aau-network-security/haaukins/blob/a8abc3a3613b5b363c3b3a41ad0e88d37ccdd5e6/daemon/proto/daemon.proto#L39) set to true or false. If it is false which means resume called in back end.

daemon/daemon.go
mrturkmen06: I think, documenting it would be enough instead of passing it through config file.

daemon/proto/daemon.proto
gianmarcomennecozzi: I see, so maybe its better to change the name of that function

mrturkmen06: What would be your proposal for the name, then :) ?

gianmarcomennecozzi: this one is okay now

mrturkmen06:
The event suspend functionality did not finished completely. The only thing is that adding suspending event into eventpool without initializing resources. [WIP]

mrturkmen06:
@gianmarcomennecozzi Would be much more easy for you to check latest changes by commit 
3104f60

mrturkmen06:
@gianmarcomennecozzi 
Note that the issue of retrieving events and arranging suspended events based on returned information from db service did not fixed.  The issue - #419 -  is pointing that problem. What this pull request includes following ; 
- Team based suspend/resume
- Event based suspend/resume 

Missing point : 

- #419 ( points for teams and events )

PR #411, opened 2020-05-27
fixed user error return

gianmarcomennecozzi:


mrturkmen06:
@gianmarcomennecozzi 
Please prefer to use `hotfix`and `feature` prefix in front of branch name.

daemon/user.go
```diff
@@ -143,7 +143,7 @@ func (d *daemon) DestroyUser(ctx context.Context, request *pb.DestroyUserRequest
 
 	if err := d.users.DeleteUserByUsername(request.Username); err != nil {
 		log.Error().Msgf("User delete error %v", err)
-		return &pb.DestroyUserResponse{Message: fmt.Sprintf("Error on deleting user %s, error: %v", request.Username, err)}, nil
+		return &pb.DestroyUserResponse{Message: fmt.Sprintf("Error on deleting user %s, error: %v", request.Username, err)}, err
```
mrturkmen06, line 146: What is the reason of changing this ?

gianmarcomennecozzi: Because we have to send back an error to the client since we got an error if its inside that `if` statement

mrturkmen06:
I think, we can merge it once branch name is changed :) @gianmarcomennecozzi

gianmarcomennecozzi:
you can merge both

Issue #421, opened 2020-06-02
Lab API error: address space


gianmarcomennecozzi:
<img width="1181" alt="83396298-0dfd1f80-a3fc-11ea-8c1b-41761f27ef4d" src="https://user-images.githubusercontent.com/55431233/83516603-b0d69c00-a4d7-11ea-95c0-8332bf6addb9.png">

PR #423, opened 2020-06-06
remove duplicate tags in event creation

mrturkmen06:

- Duplicated tags from incoming create event request deleted.

### DIFF
--- a/daemon/event.go
+++ b/daemon/event.go
@@ -46,9 +46,10 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 		Str("finishTime", req.FinishTime).
 		Msg("create event")
 	now := time.Now()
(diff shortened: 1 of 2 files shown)


mrturkmen06:
There is no relationship between #421 and the changes in this pull request @gianmarcomennecozzi

mrturkmen06:
Why did you think so ?

gianmarcomennecozzi:
yeah thats true but to solve that problem i think its good to have that for loop anyway

mrturkmen06:
> yeah thats true but to solve that problem i think its good to have that for loop anyway

I could not understand the reason of it,  could you a little bit explain @gianmarcomennecozzi ?

mrturkmen06:
@gianmarcomennecozzi, I did not understand your point, but anyway, if you would like to explain, I am always available. 

I am merging changes.

Issue #417, opened 2020-05-28
booking functionality into new version


mrturkmencom:
Haaukins with amigo is missing implementation of booking events. Nice to include booking functionality with timers maybe.

mrturkmencom:
Done with  b6be1a5

PR #425, opened 2020-06-12
Feature/integrate booking func #417

mrturkmencom:
- Event booking added 
- Event will be checked in 8 hours intervals which can be discussed to change.


mrturkmen06:
I am going to include, destroy event in scheduler function, to close events which has overdue finish time.

daemon/daemon.go
```diff
@@ -403,8 +395,52 @@ func (d *daemon) Run() error {
...
+	// initialize schedulers
+	if err := d.initializeScheduler(); err != nil {
+		return err
 	}
+
 	return s.Serve(lis)
 }
+
+// calculateTotalConsumption will add up all running events resources
+func calculateTotalConsumption(d *daemon) int {
```
gianmarcomennecozzi, line 408: this way `func (d *daemon) CalculateTotalConsumption() int {` looks better

daemon/daemon.go
```diff
@@ -403,8 +395,52 @@ func (d *daemon) Run() error {
...
+			totalConsumption += config.Capacity
+		}
+		// note that this is very raw calculation
+		// suspended resources could be excluded
+		totalConsumption += config.Available + len(e.GetTeams())
+	}
+	return totalConsumption
+}
+
+func (d *daemon) initializeScheduler() error {
```
gianmarcomennecozzi, line 422: Create a test for this function

daemon/event.go
```diff
@@ -65,19 +69,31 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 
 	finishTime, _ := time.Parse("2006-01-02", req.FinishTime)
 
+	// handling booked events might be changed
+	startTime, _ := time.Parse(time.RFC3339, req.StartTime)
```
gianmarcomennecozzi, line 73: why do we have so many different time formats? (time.RFC3339, "2006-01-02", displayTimeFormat). can we use only one or the entire repo?

daemon/event.go
```diff
@@ -111,6 +127,34 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 		return err
 	}
 	d.startEvent(ev)
+
+	return nil
+}
+
+func (d *daemon) bookEvent(ctx context.Context, req *pb.CreateEventRequest) error {
```
gianmarcomennecozzi, line 134: Do we have to create a method in the webclient for this function right?

daemon/event.go
```diff
@@ -254,3 +298,95 @@ func removeDuplicates(exercises []string) []string {
...
+	return nil
+}
+
+func (d *daemon) generateEventConfig(event *pbc.GetEventResponse_Events, status int32) store.EventConfig {
+
+	var instanceConfig []store.InstanceConfig
+	var exercises []store.Tag
+
+	requestedStartTime, _ := time.Parse(time.RFC3339, event.StartedAt)
+	requestedFinishTime, _ := time.Parse(displayTimeFormat, event.ExpectedFinishTime)
```
gianmarcomennecozzi, line 331: why two different time format here?

daemon/event.go
```diff
@@ -254,3 +298,95 @@ func removeDuplicates(exercises []string) []string {
 	}
 	return uniqueExercises
 }
+
+// a method which will start booked events by checking them in some predefined time intervals
+func (d *daemon) visitBookedEvents() error {
```
gianmarcomennecozzi, line 303: i think would be nice to have some tests for this function too

lab/hub.go
```diff
@@ -49,6 +49,7 @@ func NewHub(ctx context.Context, creator Creator, buffer int, cap int) (*hub, er
 		ctx := context.Background()
 		for range ready {
 			wg.Add(1)
+			// todo: handle this in case of error
```
gianmarcomennecozzi, line 52: this is a critical error  that has to be solved. it will cause haaukins crash in some cases

daemon/event.go
mrturkmen06: Hmm, for booking events it is a little bit old fashion, I mean, like when you do not care hour and minutes then, the likelihood of having `start event` order at same time is very high which might have negative effect on server side. Like imagine a scenario where three different event booked for  date `2020-07-10`, actually it means following ; 

`2020-07-10 00:00:00`  - for all events which are booked for same date. 

Then event scheduler will check events and see all three, will try to start time at the same time, might have negative effect on server load. 

Instead, I thought that would be good to consider date with specific time, this way the likelihood of having colliding jobs will be lower than before, 

`2020-06-20T12:41:45.14Z`  when  time format is like this. 

I hope, I managed to explain the reason why I preferred to use `time.RFC3339`, yeah true, we might use it in any time parsing operation. 

@gianmarcomennecozzi

daemon/event.go
mrturkmen06: No, I do not think, we have to.

daemon/event.go
mrturkmen06: There is no particular reason.

lab/hub.go
mrturkmen06: however, it is NOT topic of this pull request or issue, hence, do not consider it for this pull request.

daemon/daemon.go
mrturkmen06: @gianmarcomennecozzi
I think, it is enough to write test function for RunScheduler function, since initializeScheduler just initiating configuration and calls RunScheduler.

mrturkmencom:
Events which are either stopped or expired will stay on database, event tag of them will be updated with unique numbers at the end. In addition, status of them will be set to 3 which states CLOSED. Example SS given below: 

Screenshot 2020-07-03 at 02 56 15

store/proto/store.proto
```diff
@@ -6,23 +6,38 @@ service Store {
...
 message GetEventStatusRequest {
     string eventTag = 1;
 }
 
+message GetEventRequest {
+    // this will indicate that whether all events requested or some by status
+    // status 0  > Running Events (which set as Running)
+    // status 1  > Suspended Events (which set as Suspended)
+    // status 2  > Booked Events (which are booked by client)
+    int32 status = 1;
```
gianmarcomennecozzi, line 38: there is also status 3. is not it ?

store/proto/store.proto
```diff
@@ -33,6 +48,11 @@ message EventStatus {
 }
 
 
+message UpdateEventTagRequest  {
+    string oldTag = 1;
+    string newTag = 2;
+}
+
```
gianmarcomennecozzi, line 55: I think we dont need it

store/proto/store.proto
mrturkmencom: forget to add here.

PR #433, opened 2020-08-03
fixed event status

gianmarcomennecozzi:
i have found another error apart from this one that i have fixed.
When haaukins either crash or is stopped with an event in `suspended` status, you are not able to `resume` that event when haaukins starts again.
You can test it either from the `cli` or the `webclient`.

### DIFF
--- a/svcs/guacamole/event.go
+++ b/svcs/guacamole/event.go
@@ -30,8 +30,8 @@ import (
 const (
 	displayTimeFormat = "2006-01-02 15:04:05"
 	Running           = State(0)
-	Booked            = State(1) // todo: will be added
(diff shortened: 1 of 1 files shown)


mrturkmencom:
> i have found another error apart from this one that i have fixed.
> When haaukins either crash or is stopped with an event in `suspended` status, you are not able to `resume` that event when haaukins starts again.
> You can test it either from the `cli` or the `webclient`.

Yeah, however the proposed change will not fix it.

gianmarcomennecozzi:
Yeah i know, with this pull request i just wanted to make work the suspend/resume functionality from the webclient

PR #434, opened 2020-08-12
Feature/add stresstest cmd #000

mrturkmencom:
To be able to make multiple requests and see how haaukins will react on demand, a command is included. Basically, it is simple post request which is used for signing up to an event. 

**Usage**

 _Note that event should be running before hand_

`hkn event load -t test -r 3` > it means that signup 3 users on event test. 

The functionality could be improved in time and additional test parameters could be included as well.


daemon/event.go
```diff
@@ -466,6 +477,86 @@ func (d *daemon) closeEvents() error {
...
+	endPoint := fmt.Sprintf(protocol + req.EventName + "." + d.conf.Host.Http + ":" + port + "/signup")
+	resp := make(chan string)
+	for i := 0; i < int(req.NumberOfTeams); i++ {
+		go func() {
+			resp <- d.postRequest(endPoint, protocol)
+		}()
+	}
+	response := <-resp
+
+	return &pb.TestEventLoadResp{SignUpResult: response}, nil
```
gianmarcomennecozzi, line 515: I am not sure this is the correct implementation. 
Since `resp` is a channel, we can have multiple input on it (the loop confirm that), so i think we need to have to send multiple response to the client. in this case we need a `stream` in the proto file and change the code a little bit. does it make sense?

mrturkmencom: Yesss, I do agree, it will be stream based ;) I am aware of it 😉

And more, I think, having time based requests could be nice as well, like having list of times 

- [1, 5, 10, 15, 20] 

and see when the problem occurs mostly in case of an event. 


Currently the problem is that, when number of users who would like to signup bigger than queue, haaukins does not wait until a new lab is created, throws no labs available on queue error. It has cons and pros that we can discuss @gianmarcomennecozzi

gianmarcomennecozzi: I got it. we can actually tests it with the number of labs present in the buffer queue andt then create a test event with an `availability` of 20 perhaps.

gianmarcomennecozzi:
It would be good if through this stress test we can  reproduce this issue https://github.com/aau-network-security/haaukins/issues/454

mrturkmencom:
> It would be good if through this stress test we can reproduce this issue #454

Yeah, when I was checking the stress test, I faced similar errors, it is basically happens when NOT(Number Of Teams) > BUFFER.

PR #438, opened 2020-08-18
fix to add no rows in result set

mrturkmencom:
- Following error is fixed. 

Screenshot 2020-08-19 at 02 06 39


mrturkmencom:
The issue is already fixed with https://github.com/aau-network-security/haaukins/pull/439/files

PR #445, opened 2020-08-21
update status of event when booked event is started

mrturkmencom:

- It is realized that creating booked event did not setup  the status of the event on database side correctly, for this reason following changes introduced

- Waiting for closing events, according to any error or  something which is strange another update could be included, for now, pull request will remain as draft one.


gianmarcomennecozzi (approved):
It should be fine with those changes. We are going to test in sec01 at the end of this week

PR #462, opened 2020-09-05
dev env deployment

mrturkmencom:
Continuous Deployment to development environment only for develop branch. Generated binaries will be deployed in following format:
hknd-<commit-id>
hkn-client-<commit-id>
Notification to slack will be sent whenever there is a new deployment.


gianmarcomennecozzi (approved):
That's cool, we really needed it :D

PR #463, opened 2020-09-06
fix reset challenge error

gianmarcomennecozzi:
fixed reset challenge error caused children challenges tag

### DIFF
--- a/svcs/amigo/amigo.go
+++ b/svcs/amigo/amigo.go
@@ -632,6 +632,7 @@ func (am *Amigo) getTeamFromRequest(w http.ResponseWriter, r *http.Request) (*st
 	return t, nil
 }
 
+//Read json from request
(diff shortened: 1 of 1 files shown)


svcs/amigo/amigo.go
```diff
@@ -721,6 +724,16 @@ func JSONEndpoint(next http.HandlerFunc) http.HandlerFunc {
 		next(w, r)
 	}
 }
+
```
mrturkmencom, line 727: I understood it, we have already discussed.However, for safety and guideline, could you add brief explanation for the function (to see it on go docs)  by starting with "getParentChallengeTag ...."?

mrturkmencom (changes requested):

mrturkmencom (approved):

PR #464, opened 2020-09-06
Feature/add flag format #449

### LINKED ISSUE #449: Report Flag Format -- Feedback from IT CPH University
- [x] Document in the Amigo home page what has to be found in each challenge, and how looks like the flag

- [x] It would be good have a common flag format to type  for example `haaukins{<flag>}` . It can be easily managed by haaukins and not much source code needs to be changed

mrturkmencom:
- HKN{} flag format introduced
- Necessary parts are changed 
Closes #449


gianmarcomennecozzi (approved):
Good, users have finally something to look for inside the challenge

PR #465, opened 2020-09-07
added general info for users

gianmarcomennecozzi:
made modal that show useful information for users such as:
- how to copy from clipboard
- how to fix connection error


mrturkmencom (approved):

kres0345:
not visible enough still, our 5 man team found out copy-paste was a thing in the last 30 minutes

mrturkmencom:
Thanks for the feedback @kres0345, we might think to create another page for information button. Existing information button is sufficient enough, according to feedback, we can re-consider to change it.

PR #489, opened 2020-09-29
Hotfix/fix static flag #000

mrturkmencom:
Issue regarding to static flags has been fixed. 
Challenge file should have complete flag if it is static: HKN{xxxxxxxxxxxx}


gianmarcomennecozzi (approved):
Good :D

PR #500, opened 2020-10-09
Revert/step by step #000

mrturkmencom:


gianmarcomennecozzi (approved):
I think it is okay. Maybe we have to regenerate the compiled proto file.

PR #525, opened 2020-10-26
Hotfix/suspend patch #000

mrturkmencom:


gianmarcomennecozzi (approved):
After the event of yesterday seems that everythings was okay, so all the fixes made to the `master` branch has been applied here as well.
Thanks man. well done

Issue #527, opened 2020-10-26
ERR Error while creating new lab API error (403)


mrturkmencom:
ERR Error while creating new lab API error (403): Pool overlaps with other one on this address space 

Given error above occurs in lab creation time to time, add try mechanism to it, in order to do not crash haaukins entirely. 

https://github.com/aau-network-security/haaukins/blob/d2bb23be398b35dad0d00a5646c45eb49864f73f/lab/hub.go#L54

mrturkmencom:
New error regarding to this issue has been seen. 
![Screenshot 2020-11-17 at 11 26 59](https://user-images.githubusercontent.com/13614433/99378615-e9bcce80-28c7-11eb-85b1-7eefe5b969a3.png)

mrturkmencom:
#551

mrturkmencom:
No error received after the patch #551

PR #529, opened 2020-10-27
add re-try mechanism to NewLab() critical error

mrturkmencom:


mrturkmencom:
Closes #527

gianmarcomennecozzi (commented):
I think we have already tested the same approach, several months ago, and i remeber that it didnt work well.
Basically if we get an error while creating the event, we will get the same error for all the other iteration in the for loop. This what we faced last time.

PR #589, opened 2021-01-27
Feature/add auth to secret challenges #000

mrtrkmn:


mrtrkmn (commented):
auth added for secret challenges to spin off

PR #617, opened 2021-03-19
Develop

mrtrkmn:
upcoming version of haaukins should use latest version of
- haaukins-exercises v1.1.7
- haaukins-store v2.1.1


mrtrkmn (commented):
- Progressbar dependency removed
- IPTable manipulation for VPN based events added
- Issue regarding to metasploit and python web server setting on VPN scenario fixed

PR #656, opened 2021-08-06
Feature/html descriptions #000

Mikkelhost:
In the challenge configs under td variable you can now write markdown in the description like so
```
td: |
  # Title
  ## Subtitle
  * Bulletpoints
  1. Ordered list
  
```
It is possible to use all ascii chars after the "|" character. See the commits on "PESUID proc" how i have tested the different markdowns


mrtrkmn:
Hey @Mikkelhost, 
It looks nice ! Could you add an image of the feature to understand how it looks like ?

PR #667, opened 2021-08-17
Hotfix/static flag environ fix #000

Mikkelhost:
Env var is no longer used when static flag is present in the config file. it will simply be ignored

### DIFF
--- a/store/exercise.go
+++ b/store/exercise.go
@@ -84,12 +84,9 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
 				Tag:   flag.Tag,
 				Value: value,
 			})
-
-			if static {
-				//Do Nothing with flag env var
-			} else {
+			if !static {
 				envVars[flag.EnvVar] = value
-			}
+			} 
 
 		}
 
--- a/store/exercise.go
+++ b/store/exercise.go
@@ -69,7 +69,7 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
 
 		for _, flag := range conf.Flags {
 			value := flag.StaticFlag
-			var static = false
+			var static = true
 			// static flag format in exercises file
 			//  should obey flag format HKN{*********}
 			if value == "" {
--- a/store/exercise.go
+++ b/store/exercise.go
@@ -69,11 +69,14 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
 
 		for _, flag := range conf.Flags {
 			value := flag.StaticFlag
+			var static = false
 			// static flag format in exercises file
 			//  should obey flag format HKN{*********}
 			if value == "" {
 				// flag is not static
 				value = NewFlag().String()
+			} else {
+				static = true
 			}
 
 			challenges = append(challenges, Challenge{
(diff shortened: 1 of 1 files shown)


store/exercise.go
```diff
@@ -69,19 +69,28 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
 
 		for _, flag := range conf.Flags {
 			value := flag.StaticFlag
+			var static = false
```
mrtrkmn, line 72:
```suggestion
			var static = true
```

store/exercise.go
```diff
@@ -69,19 +69,28 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
...
 				Tag:   flag.Tag,
 				Value: value,
 			})
-			envVars[flag.EnvVar] = value
+
+			if static {
+				//Do Nothing with flag env var
+			} else {
+				envVars[flag.EnvVar] = value
+			}
```
mrtrkmn, line 92:
```suggestion
			if !static {
				envVars[flag.EnvVar] = value
			} 
```

store/exercise.go
```diff
@@ -69,19 +69,28 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
...
 			value := flag.StaticFlag
+			var static = false
 			// static flag format in exercises file
 			//  should obey flag format HKN{*********}
 			if value == "" {
 				// flag is not static
 				value = NewFlag().String()
+			} else {
+				static = true
 			}
```
mrtrkmn, line 80:
```suggestion
			if value == "" {
				// flag is not static
				value = NewFlag().String()
				static = false
			} 
```

mrtrkmn (changes requested):
Hey Mikkel, 
I have refactored it a little bit, could you test the changes and re-push ?  
( -btw, you can also directly accept suggested changes, and it will commit directly -)

store/exercise.go
```diff
@@ -69,19 +69,24 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
 
 		for _, flag := range conf.Flags {
 			value := flag.StaticFlag
+			var static = true
```
mrtrkmn, line 72:
```suggestion
			
```

store/exercise.go
```diff
@@ -69,19 +69,24 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
...
 			// static flag format in exercises file
 			//  should obey flag format HKN{*********}
 			if value == "" {
 				// flag is not static
 				value = NewFlag().String()
-			}
+				static = false
```
mrtrkmn, line 78:
```suggestion
				envVars[flag.EnvVar] = value
```

store/exercise.go
```diff
@@ -69,19 +69,24 @@ func (e Exercise) ContainerOpts() []ContainerOptions {
...
 
 			challenges = append(challenges, Challenge{
 				Name:  flag.Name,
 				Tag:   flag.Tag,
 				Value: value,
 			})
-			envVars[flag.EnvVar] = value
+			if !static {
+				envVars[flag.EnvVar] = value
+			}
```
mrtrkmn, line 88:
```suggestion
			
```

mrtrkmn (changes requested):

mrtrkmn (approved):

PR #675, opened 2021-08-30
Can now save and list profiles from db, just need the rest

Mikkelhost:
Screenshot from 2021-08-30 02-28-07


daemon/event.go
```diff
@@ -846,3 +847,143 @@ func getVPNIP() (string, error) {
...
+		Info().
+		Str("name", req.Name).
+		Msg("Deleting profile")
+	log.Info().Str("profileName", req.Name).Msg("Trying to delete profile")
+	if !user.NPUser {
+		_, err := d.dbClient.DeleteProfile(ctx, &pbc.DelProfileRequest{
+			Name: req.Name,
+		})
+		if err != nil {
+			return fmt.Errorf("Error when updating profile: %e", err)
```
mrtrkmn, line 983:
```suggestion
			return fmt.Errorf("Error when deleting profile: %e", err)
```

daemon/event.go
```diff
@@ -846,3 +847,143 @@ func getVPNIP() (string, error) {
...
+		return fmt.Errorf("user credentials could not found on context %v", err)
+	}
+	log.Ctx(ctx).
+		Info().
+		Str("name", req.Name).
+		Msg("Saving profile")
+	log.Info().Str("profileName", req.Name).Msg("Trying to save profile")
+	var challenges []*pbc.AddProfileRequest_Challenge
+	for _, c := range req.Challenges {
+		//log.Info().Str("tag", c.Tag).Str("name", c.Name).Msg("Appending challenge to profile")
```
mrtrkmn, line 865: If not used, it would be nice to remove the commented codes

daemon/event.go
```diff
@@ -846,3 +847,143 @@ func getVPNIP() (string, error) {
...
+
+	profilesStore, err := d.dbClient.GetProfiles(ctx, &pbc.EmptyRequest{})
+	if err != nil {
+		return &pb.ListProfilesResponse{}, fmt.Errorf("[store-service]: ERR getting profiles %v", err)
+	}
+	var profs []store.Profile
+	for _, p := range profilesStore.Profiles {
+		profile, err := protobufToJson(p)
+		if err != nil {
+			//log.Info().Msgf("Error in protobufToJson")
```
mrtrkmn, line 903: If not used, it would be nice to remove the commented codes

daemon/event.go
```diff
@@ -846,3 +847,143 @@ func getVPNIP() (string, error) {
...
+			return nil, err
+		}
+		pstruct := store.Profile{}
+		json.Unmarshal([]byte(profile), &pstruct)
+		profs = append(profs, pstruct)
+		//log.Info().Msgf("Got profile: %s", profile)
```
mrtrkmn, line 909: If not used, it would be nice to remove the commented codes

mrtrkmn (changes requested):
That's great ! 
It would be nice to remove codes which are commented and not used. Then it is ready to go

mrtrkmn (approved):

PR #692, opened 2021-11-17
Added dynamic docker user on guacd container

Mikkelhost:
AS haaukins is not run as root on sec02 and sec01, the guacd container which contains the shared files needed to be run with the same uid and gid as the user running haaukins. This should be fixed with this PR

### DIFF
--- a/svcs/guacamole/guacamole.go
+++ b/svcs/guacamole/guacamole.go
@@ -165,7 +165,7 @@ func (guac *guacamole) create(ctx context.Context, eventTag string) error {
 		Mounts: []string{
 			vbox.FileTransferRoot + "/" + eventTag + "/:/home/",
 		},
-		User: uid + ":" + gid,
+		User: user
 	})
 
 	mysqlPass := uuid.New().String()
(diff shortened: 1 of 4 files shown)


Lanestolen (commented):
Thing i think would make the code a bit prettier, take it or leave it :)

svcs/guacamole/guacamole.go
```diff
@@ -146,10 +148,13 @@ func (guac *guacamole) GetAdminPass() string {
 
 //TODO choose another path for mount, Create new path when making a new event.
 func (guac *guacamole) create(ctx context.Context, eventTag string) error {
-	err := vbox.CreateEventFolder(eventTag)
-	if err != nil {
-		return err
+	_ := vbox.CreateEventFolder(eventTag)
+		// Do nothing
 	}
```
Mikkelhost, line 153:
```suggestion
	
```

Mikkelhost:
Tests on sec01 shows promising however.
Files does not remove if an event naturally stops because it reaches it's end.
Therefore this needs to be added to this hotfix

PR #703, opened 2022-05-04
Hotfix/fix start event issue #701

### LINKED ISSUE #701: events are saved to db even though it failed to start
**Describe the bug**
The event is saved to database even though it is not successfully created. It can lead some problems when the platform is back, there should not be manual manipulation of database. 

**To Reproduce**
Call create event with some parameters which is actually not present. Anything which causes the platform to crash. 

**Expected behavior**

The platform should handle this issue, like when there is a problem of creating event, all changes (if any) should be reverted. 

**Preliminary check is required**

mrtrkmn:
Closes #701 

- Checking available frontends before actually starting to spin up resources 
- Remove CreateEventFromDB function, add SaveToDB. 
- CreateEventFromDB function was not performing what it says.

### DIFF
--- a/daemon/daemon_test.go
+++ b/daemon/daemon_test.go
@@ -290,6 +290,7 @@ func (c *counter) Value() (val int32) {
 // RunScheduler runs multiple goroutines
 // in different time intervals
 func TestRunScheduler(t *testing.T) {
+	t.Skipf("Race condition is happenning in this particular test --> needs to be checked")
 	actualC1 := counter{}
 	actualC2 := counter{}
 
--- a/daemon/event.go
+++ b/daemon/event.go
@@ -5,15 +5,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/aau-network-security/haaukins/virtual/vbox"
 	"math"
 	"math/rand"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/aau-network-security/haaukins/virtual/vbox"
 
 	pb "github.com/aau-network-security/haaukins/daemon/proto"
 	eproto "github.com/aau-network-security/haaukins/exercise/ex-proto"
@@ -92,6 +94,34 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 		return fmt.Errorf("user credentials could not found on context %v", err)
 	}
 
+	// checking through eventPool is not good enough since booked events are not added to eventpool
+	isEventExist, err := d.dbClient.IsEventExists(ctx, &pbc.GetEventByTagReq{
+		EventTag: req.Tag,
+		// it will take INVERT condition which means that query from
+		//Running, Suspended and Booked events
+		Status: Closed,
+	})
+	if err != nil {
+		noEventExist := fmt.Errorf("event does not exist or something is wrong: %v", err)
+		resp.Send(&pb.LabStatus{ErrorMessage: noEventExist.Error()})
+		return noEventExist
+	}
+	if isEventExist.IsExist {
+		resp.Send(&pb.LabStatus{ErrorMessage: NotAvailableTag})
+		return fmt.Errorf(NotAvailableTag)
+	}
+	log.Debug().Msgf("Checked existing events through database.")
+
+	// preliminary check for provided frontend
+	// no need to check for VPN access
+	if req.OnlyVPN != int32(VPN) || req.OnlyVPN != int32(VPNBrowser) {
+		availableFrontendsMap := d.frontends.GetAllAvailableFrontends()
+		_, ok := availableFrontendsMap[req.Frontends[0]]
+		if !ok {
+			return fmt.Errorf("\nProvided frontend [ %s ] cannot be found ! \nAvailable frontends are:  %v ", req.Frontends[0], reflect.ValueOf(availableFrontendsMap).MapKeys())
+		}
+	}
+
 	if (req.OnlyVPN == VPN || req.OnlyVPN == VPNBrowser) && req.Capacity > 253 {
 		resp.Send(&pb.LabStatus{ErrorMessage: CapacityExceedsErr.Error()})
 		return CapacityExceedsErr
@@ -147,23 +177,6 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 			return invalidDateErr
 		}
 
-		// checking through eventPool is not good enough since booked events are not added to eventpool
-		isEventExist, err := d.dbClient.IsEventExists(ctx, &pbc.GetEventByTagReq{
-			EventTag: req.Tag,
-			// it will take INVERT condition which means that query from
-			//Running, Suspended and Booked events
-			Status: Closed,
-		})
-		if err != nil {
-			noEventExist := fmt.Errorf("event does not exist or something is wrong: %v", err)
-			resp.Send(&pb.LabStatus{ErrorMessage: noEventExist.Error()})
-			return noEventExist
-		}
-		if isEventExist.IsExist {
-			resp.Send(&pb.LabStatus{ErrorMessage: NotAvailableTag})
-			return fmt.Errorf(NotAvailableTag)
-		}
-		log.Debug().Msgf("Checked existing events through database.")
 		// difference  in days
 		// if there is no difference in days it means event  will be
 		// started immediately
@@ -237,7 +250,10 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 			resp.Send(&pb.LabStatus{ErrorMessage: err.Error()})
 			return err
 		}
-
+		if err := d.ehost.SaveEventToDB(ctx, conf); err != nil {
+			ev.Close()
+			return err
+		}
 		d.startEvent(ev)
 	}
 	return nil
@@ -483,13 +499,15 @@ func (d *daemon) ListEvents(ctx context.Context, req *pb.ListEventsRequest) (*pb
 	return &pb.ListEventsResponse{Events: events}, nil
 }
 
+// does not call SaveToDB function
+// creates event from config
+// where config retrieved from DB
 func (d *daemon) createEventFromEventDB(ctx context.Context, conf store.EventConfig) error {
 
 	if err := conf.Validate(); err != nil {
 		return err
 	}
-
-	ev, err := d.ehost.CreateEventFromEventDB(ctx, conf, d.conf.Rechaptcha)
+	ev, err := d.ehost.CreateEventFromConfig(ctx, conf, d.conf.Rechaptcha)
 	if err != nil {
 		log.Error().Err(err).Msg("Error creating event from database event")
 		return err
--- a/daemon/event.go
+++ b/daemon/event.go
@@ -114,7 +114,7 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
 
 	// preliminary check for provided frontend
 	// no need to check for VPN access
-	if req.OnlyVPN != int32(VPN) || req.OnlyVPN != int32(VPNBrowser) {
+	if req.OnlyVPN == int32(NoVPN) || req.OnlyVPN == int32(VPNBrowser) {
 		availableFrontendsMap := d.frontends.GetAllAvailableFrontends()
 		_, ok := availableFrontendsMap[req.Frontends[0]]
 		if !ok {
--- a/daemon/event.go
+++ b/daemon/event.go
@@ -9,7 +9,6 @@ import (
 	"math/rand"
 	"net/http"
 	"net/url"
-	"reflect"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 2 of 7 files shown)


Mikkelhost:
When you are saying checking for available frontends?
Does it check if it has been configured?

I have been experiencing that the webclient displays all files which ends with .ova

The expected behavior i would like is that the webclient only displays the vm's that has been configured in frontends.yml

Mikkelhost:
From my experience, if you try to use a vm that has not been configured in the frontends.yml file it tries to start the vm with 0 memory. Vbox comes back with an error on that one which in turn makes haaukins crash with a nil pointer de-reference panic. Either i would suggest to only try and start the vm if the vm has been configured in the frontends.yml or i would make the function that sets the vm memory or whatever, check if it is 0. If it is 0, it should just default to 4GB or some other default value.

mrtrkmn:
> From my experience, if you try to use a vm that has not been configured in the frontends.yml file it tries to start the vm with 0 memory. Vbox comes back with an error on that one which in turn makes haaukins crash with a nil pointer de-reference panic. Either i would suggest to only try and start the vm if the vm has been configured in the frontends.yml or i would make the function that sets the vm memory or whatever, check if it is 0. If it is 0, it should just default to 4GB or some other default value.

GetAllAvailableFrontends function gets infromation from the frontends.yml file, so from already configured place. However, as you suggested it would be ideal to set default memory if memory is not set correctly, also check if ova file is exists.

Mikkelhost:
> > From my experience, if you try to use a vm that has not been configured in the frontends.yml file it tries to start the vm with 0 memory. Vbox comes back with an error on that one which in turn makes haaukins crash with a nil pointer de-reference panic. Either i would suggest to only try and start the vm if the vm has been configured in the frontends.yml or i would make the function that sets the vm memory or whatever, check if it is 0. If it is 0, it should just default to 4GB or some other default value.
> 
> GetAllAvailableFrontends function gets infromation from the frontends.yml file, so from already configured place. However, as you suggested it would be ideal to set default memory if memory is not set correctly, also check if ova file exists.

Yes!

mrtrkmn:
Roger, I am going to add them.

mrtrkmn:
On top of changes done earlier, default values for frontend image is added and checked during creating an event with last commit. 
What are your feedback on this PR @Lanestolen and @Mikkelhost  ?

daemon/daemon_test.go
```diff
@@ -290,6 +290,7 @@ func (c *counter) Value() (val int32) {
 // RunScheduler runs multiple goroutines
 // in different time intervals
 func TestRunScheduler(t *testing.T) {
+	t.Skipf("Race condition is happenning in this particular test --> needs to be checked")
```
Lanestolen, line 293: Maybe add an issue so this isn't forgotten

daemon/event.go
```diff
@@ -92,6 +94,34 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
...
+	if err != nil {
+		noEventExist := fmt.Errorf("event does not exist or something is wrong: %v", err)
+		resp.Send(&pb.LabStatus{ErrorMessage: noEventExist.Error()})
+		return noEventExist
+	}
+	if isEventExist.IsExist {
+		resp.Send(&pb.LabStatus{ErrorMessage: NotAvailableTag})
+		return fmt.Errorf(NotAvailableTag)
+	}
+	log.Debug().Msgf("Checked existing events through database.")
```
Lanestolen, line 113:
```suggestion
	log.Debug().Msgf("Checked existing events through database.")
	log.Debug().Msg("Checked existing events through database.")
```
Maybe add additional fields for the logging like which tag was used when checking

daemon/event.go
```diff
@@ -92,6 +94,34 @@ func (d *daemon) CreateEvent(req *pb.CreateEventRequest, resp pb.Daemon_CreateEv
...
+		Status: Closed,
+	})
+	if err != nil {
+		noEventExist := fmt.Errorf("event does not exist or something is wrong: %v", err)
+		resp.Send(&pb.LabStatus{ErrorMessage: noEventExist.Error()})
+		return noEventExist
+	}
+	if isEventExist.IsExist {
+		resp.Send(&pb.LabStatus{ErrorMessage: NotAvailableTag})
+		return fmt.Errorf(NotAvailableTag)
```
Lanestolen, line 111: Why the format?

lab/lab.go
```diff
@@ -138,13 +140,20 @@ func (l *lab) addFrontend(ctx context.Context, conf store.InstanceConfig, rdpPor
 	if err != nil {
 		return nil, err
 	}
-
+	var mem uint
+	if conf.MemoryMB <= 0 || conf.MemoryMB < defaultImageMEMMB/2 {
+		log.Debug().Msgf("Memory cannot be smaller or equal to zero or less than [ %d ], setting it to default value [ %d ] ", defaultImageMEMMB/2, defaultImageMEMMB)
+		mem = defaultImageMEMMB
+		log.Warn().Msgf("Check frontends configuration file to update memory value of image [ %s ] ", conf.Image)
```
Lanestolen, line 147: Use fields instead of the formatting?

lab/lab.go
```diff
@@ -138,13 +140,20 @@ func (l *lab) addFrontend(ctx context.Context, conf store.InstanceConfig, rdpPor
 	if err != nil {
 		return nil, err
 	}
-
+	var mem uint
+	if conf.MemoryMB <= 0 || conf.MemoryMB < defaultImageMEMMB/2 {
+		log.Debug().Msgf("Memory cannot be smaller or equal to zero or less than [ %d ], setting it to default value [ %d ] ", defaultImageMEMMB/2, defaultImageMEMMB)
```
Lanestolen, line 145: Is there a reason why memory should be higher than 2gb?, also see commend bellow

svcs/guacamole/event.go
```diff
@@ -95,8 +95,55 @@ type eventHost struct {
 	dir       string
 }
 
-//Create the event configuration for the event got from the DB
-func (eh *eventHost) CreateEventFromEventDB(ctx context.Context, conf store.EventConfig, reCaptchaKey string) (Event, error) {
+func protobufToJson(message proto.Message) (string, error) {
```
Lanestolen, line 98: Why not proto directly to struct?

virtual/vbox/vbox.go
```diff
@@ -392,6 +394,13 @@ func (lib *vBoxLibrary) getPathFromFile(file string) string {
 
 func (lib *vBoxLibrary) GetCopy(ctx context.Context, conf store.InstanceConfig, vmOpts ...VMOpt) (VM, error) {
 	path := lib.getPathFromFile(conf.Image)
+	// check whether provided image exists or not
+	// if not exits use default image kali.ova
+	if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
+		// image does not exists, continue with default image
+		log.Debug().Msgf("Requested image [ %s ] cannot be found, continuing with default image [ %s ] ", conf.Image, defaultImage)
```
Lanestolen, line 401: Use fields in logging
Why have a default image? if no correct image send error and let user try again?

Lanestolen (changes requested):

svcs/guacamole/event.go
mrtrkmn: protobufToJson did not added on this PR, it was there, so it would be nice to consider it separately in another discussion.

virtual/vbox/vbox.go
mrtrkmn: I made some assumptions when doing it. Like I thought that `kali.ova` file is always exists, since it is as for as I know. However, that's right that assumptions are evil in most of the cases. ListFrontends function should list frontends which are configured on frontends.yml and under ova dir together. 

Furthermore, there might be the cases where ova file is corrupted and cannot be imported into the platform environment. In this case, there is a need to check whether provided ova file is corrupted or not. However, this issue itself might include more changes than it should. Therefore, my suggestion for integrity checking would be to do it in separate issue. 

So, currently if no correct image send > then kali ova will be chosen however if there is no kali.ova or corrupted one, then platform might crash. But this is another issue which is not directly related to this issue. 

This PR meant to solve issues like ;

- when create event  cmd send and there is no image configured as given from user then event information should not written to database until event is successfully up and running
- when non-exist frontend chosen, the platform should prefer default one > kali.ova
- when RAM/CPU values are not set correctly on frontends config file, the platform should prefer default values > 4096 MB , 1 CPU 

Of course, this PR is not cure to all issues it may happen but it heals some obvious ones.

lab/lab.go
mrtrkmn: Yes, there is a reason actually, there are some resource intensive tools such as Wireshark, Burpsuite and etc, those require more resources to be used in good condition. As for as I know, there are some exercises which require to use some of the mentioned tools above.

daemon/event.go
mrtrkmn: Again, these logs are not added in this PR. In my opinion, different issues on the platform should not be touched on a PR which is meant to fix particular issue.

daemon/event.go
mrtrkmn: This is not done in this PR, there are general inconsistencies of error handling on the platform itself. Can be handled with a separate issue.

lab/lab.go
Lanestolen: Fair, I just wanted the reasoning for setting it to 2gb

daemon/event.go
Lanestolen: Okay, if you want it to be handled in another issue i think it should be created

daemon/event.go
Lanestolen: Totally fair i just believe that it would be nice to improve the function as a whole when fixing issues caused but it :)

daemon/event.go
mrtrkmn: Arranging error handling guideline would be preferred way to go in my opinion. Like we should use `fmt.Errorf` in these particular situation otherwise `errors.New` and etc. Your proposal an issue would be nice to see @Lanestolen :)

virtual/vbox/vbox.go
Lanestolen: Im just trying to ensure that no inconvenient errors will happen if someone want to spin up their own instance of haaukins, it is totally right that at our deployment the kali.ova exist but it might not in other places.

I know this is not a part of the issue that this PR is trying to solve, however these changes is included here and therefore I chose to evaluate it as a part of this PR

daemon/daemon_test.go
mrtrkmn: Roger !

virtual/vbox/vbox.go
mrtrkmn: Just changed, so when provided frontend is not available on user side, program will check available frontends if there are any ova file available, platform will continue with first available file. 
- If provided ova file cannot be found go for first available frontend
- If frontend did not configured on frontends file, do not produce error instead run it with default values.
What are your feedbacks @Lanestolen & @Mikkelhost ?

mrtrkmn:
I am thinking to merge these changes @Mikkelhost  and @Lanestolen. Just for your information.

Lanestolen:
If @Mikkelhost approves it is fine with me

Mikkelhost:
Looks good to me, i agree on the points stated by @Lanestolen 
Making assumptions is a nono. 

I would just state, that instead of creating an event with a vm which the user did not mean to use, i would respond with an error instead to the user.
Lets take an example:
1. I would like to create an event with 100 capacity and 100 availability like we do for the DDC.
2. I somehow select a vm which is not available.
3. I start the whole event which at times easily takes 30 minutes with this capacity.
4. I discover the vm is wrong, so i now have to shut down the event (30 min - 60 min) to create a new one with the right vm.

Hence i would much rather have it give me an error if i do not specify the right vm or in general if i specify any wrong parameter :)

mrtrkmn:
> Looks good to me, i agree on the points stated by @Lanestolen Making assumptions is a nono.
> 
> I would just state, that instead of creating an event with a vm which the user did not mean to use, i would respond with an error instead to the user. Lets take an example:
> 
> 1. I would like to create an event with 100 capacity and 100 availability like we do for the DDC.
> 2. I somehow select a vm which is not available.
> 3. I start the whole event which at times easily takes 30 minutes with this capacity.
> 4. I discover the vm is wrong, so i now have to shut down the event (30 min - 60 min) to create a new one with the right vm.
> 
> Hence i would much rather have it give me an error if i do not specify the right vm or in general if i specify any wrong parameter :)

Roger, will look into into !

mrtrkmn:
The thing is that, it is not possible to create event with non-existing image through webclient interface, since it only lists ova files which are exist. However, it does not mean, it cannot be created, either using CLI or basic curl requests it is possible for this reason last commit checks image path if does not exists, returns an error. 

nonex

Mikkelhost:
> The thing is that, it is not possible to create event with non-existing image through webclient interface, since it only lists ova files which are exist. However, it does not mean, it cannot be created, either using CLI or basic curl requests it is possible for this reason last commit checks image path if does not exists, returns an error.
> 
> nonex
> 
> @Mikkelhost

This will also cancel event creation right?
It was mainly a concern of user friendliness, all i want to happen is if i try to create an event, and i use values which i expect to work, then i would rather get an error and have to try again than it creating an event for me with some default values :)

mrtrkmn:
> This will also cancel event creation right?

Yes, it is just returning error without initializing anything

Mikkelhost:
Then i think it's looking good :)

PR #706, opened 2022-05-04
upgrade dependencies based on security offers

mrtrkmn:
Fixes Security Issues
Tested


Lanestolen (approved):
I see no problems with upgrading these, tested it as well no problems :)

Mikkelhost:
As long as it has been thoroughly tested that none of the dependencies break anything :)

mrtrkmn:
The problem when they do not serve very well, for instance some of the dependencies reported high usage of CPU which was also an issue on Haaukins platform. It may have a link or not, but it is better to update dependencies since they are improving bugs/issues and other important features in their repository as well. 

I have tested functionalities and common points on Haaukins after update, it did not complain anything, I might miss something but this is why Haaukins has development/staging server as well. What I think is that, it would be nice to run it for some time, on staging server to extensive tests if you think it is required.

mrtrkmn:
I am thinking to merge the changes to develop. What are your feedbacks @Lanestolen and @Mikkelhost  ?

PR #711, opened 2022-05-17
update workflow action versions

### LINKED ISSUE #709: update versions on workflow files
Currently, workflow files contain outdated versions, we need to have up-to-date actions to be more in safe side.

mrtrkmn:
Closes #709

### DIFF
--- a/.github/workflows/checkbrname.yml
+++ b/.github/workflows/checkbrname.yml
@@ -11,9 +11,9 @@ jobs:
   formalities:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - name: Set up Python 3.7
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v3
         with:
           python-version: 3.7
       - name: Extract branch name
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,15 +20,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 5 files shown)


.github/workflows/checkbrname.yml
```diff
@@ -11,9 +11,9 @@ jobs:
   formalities:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - name: Set up Python 3.7
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v3
```
Lanestolen, line 16: Does this action even work? 
Isn't it supposed to add the PRs to the issues?

.github/workflows/test.yml
```diff
@@ -20,15 +20,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ '1.15', '1.16' ]
+        go: [ '1.17', '1.18' ]
```
Lanestolen, line 23: Is this reflected in the readme that haaukins only runs on 1.17 and up?

mrtrkmn: it is not reflected, documentation update should come, like
> Working state of this platform is tested on following go versions : [1.17, 1.18] 
> The platform may or may not work for other Go versions. Please take care it at your own risks

.github/workflows/checkbrname.yml
mrtrkmn: Oh I realized just now, seems it was all inactive but dangling there. This is strange.

Lanestolen (approved):

PR #717, opened 2022-06-02
update code based on updated config file

### LINKED ISSUE #716: re-format main configuration file
It is not a best practice to see some port numbers bind into binary. 
It would be ideal to get all values which might be changed env to env from a configuration file. 
This issue is opened by considering this gRPC port value
https://github.com/aau-network-security/haaukins/blob/c4e1f948bb496efaa9cfca1f517b104a6ffec2e6/daemon/daemon.go#L58
 but this is not limited to only that value, anything which is better to take it from config file should be introduced.

mrtrkmn:
Closes #716

### DIFF
--- a/daemon/config.go
+++ b/daemon/config.go
@@ -10,13 +10,16 @@ import (
 
 type Config struct {
 	Host struct {
-		Http string `yaml:"http,omitempty"`
-		Grpc string `yaml:"grpc,omitempty"`
+		Http struct {
+			Endpoint string `yaml:"endpoint,omitempty"`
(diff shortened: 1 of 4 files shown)


mrtrkmn:
The host part in main configuration file of Haaukins project needs to be updated with following:

```yaml

host:
  http:
    endpoint: localhost
    port:
      insecure: 8080
      secure: 8081
  grpc:
    endpoint: localhost
    port: 5454
```

Mikkelhost:
Looks fine!

PR #721, opened 2022-06-09
Changed markdown parser to as markdown didnt match gitlab. Apperently…

Mikkelhost:
Changed markdown renderer

### DIFF
--- a/daemon/exercise.go
+++ b/daemon/exercise.go
@@ -1,6 +1,7 @@
 package daemon
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -8,16 +9,17 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/yuin/goldmark/renderer/html"
+
 	pb "github.com/aau-network-security/haaukins/daemon/proto"
 	eproto "github.com/aau-network-security/haaukins/exercise/ex-proto"
 	"github.com/aau-network-security/haaukins/store"
 	storeProto "github.com/aau-network-security/haaukins/store/proto"
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
-	"github.com/gomarkdown/markdown"
-	"github.com/gomarkdown/markdown/parser"
 	"github.com/microcosm-cc/bluemonday"
 	"github.com/rs/zerolog/log"
+	"github.com/yuin/goldmark"
 )
 
 func (d *daemon) ListCategories(ctx context.Context, req *pb.Empty) (*pb.ListCategoriesResponse, error) {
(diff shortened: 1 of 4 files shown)


daemon/exercise.go
```diff
@@ -45,11 +47,16 @@ func (d *daemon) ListCategories(ctx context.Context, req *pb.Empty) (*pb.ListCat
...
 		md := []byte(c.CatDescription)
-		unsafeHtml := markdown.ToHTML(md, parser, nil)
+		var buf bytes.Buffer
+		renderer := goldmark.New(
+			goldmark.WithRendererOptions(html.WithUnsafe()),
+		)
+		if err := renderer.Convert(md, &buf); err != nil {
+			log.Error().Msgf("Error converting to commonmark: %s", err)
+		}
+		unsafeHtml := buf.Bytes()
```
mrtrkmn, line 59: let's make this part as function and just call it.

mrtrkmn (changes requested):

daemon/exercise.go
Mikkelhost: Haha im so stupid, i'll do that as soon i get time! :D

daemon/exercise.go
```diff
@@ -45,11 +47,16 @@ func (d *daemon) ListCategories(ctx context.Context, req *pb.Empty) (*pb.ListCat
...
-		parser := parser.NewWithExtensions(extensions)
 
 		md := []byte(c.CatDescription)
-		unsafeHtml := markdown.ToHTML(md, parser, nil)
+		var buf bytes.Buffer
+		renderer := goldmark.New(
+			goldmark.WithRendererOptions(html.WithUnsafe()),
+		)
+		if err := renderer.Convert(md, &buf); err != nil {
+			log.Error().Msgf("Error converting to commonmark: %s", err)
```
Lanestolen, line 57:
```suggestion
			log.Error().Err(err).Msg("converting to markdown")
```
Use the fields of the log package, might be nice if decided to do central logging for the distributed version later on as it makes it nicer to filter logs

daemon/exercise.go
```diff
@@ -121,11 +128,15 @@ func (d *daemon) ListExercises(ctx context.Context, req *pb.Empty) (*pb.ListExer
...
-		parser := parser.NewWithExtensions(extensions)
-
 		md := []byte(e.OrgDescription)
-		unsafeHtml := markdown.ToHTML(md, parser, nil)
+		var buf bytes.Buffer
+		renderer := goldmark.New(
+			goldmark.WithRendererOptions(html.WithUnsafe()),
+		)
+		if err := renderer.Convert(md, &buf); err != nil {
+			log.Error().Msgf("Error converting to commonmark: %s", err)
```
Lanestolen, line 137: See my other comments

svcs/amigo/table.go
```diff
@@ -184,11 +187,15 @@ func (fd *FrontendData) initChallenges(teamId string) []byte {
...
-		parser := parser.NewWithExtensions(extensions)
-
 		md := []byte(r.ChalInfo.TeamDescription)
-		unsafeHtml := markdown.ToHTML(md, parser, nil)
+		var buf bytes.Buffer
+		renderer := goldmark.New(
+			goldmark.WithRendererOptions(html.WithUnsafe()),
+		)
+		if err := renderer.Convert(md, &buf); err != nil {
+			log.Error().Msgf("Error converting to commonmark: %s", err)
```
Lanestolen, line 196: Again log and errors shouldn't start with error

Lanestolen (changes requested):

eyJhb:
Reconsider if it makes sense to use `bluemonday` at all, since you're already taking "trusted" input, ie. not user specified input.
Doesn't seem to make sense to use Goldmark with unsafe HTML, and then try to sanitize it afterwards.

Mikkelhost:
> Reconsider if it makes sense to use `bluemonday` at all, since you're already taking "trusted" input, ie. not user specified input. Doesn't seem to make sense to use Goldmark with unsafe HTML, and then try to sanitize it afterwards.

I think when i created this i pretty much went from a zero trust perspective. And i think it is a good idea to keep it as a zero trust. For the goldmark with unsafe html, i will take a look, i havn't really done anything else thank making it work :)

PR #728, opened 2022-08-09
fix returned error message and log it properly

### LINKED ISSUE #723: returns false error message on event creation
**Describe the bug**
When creating an event on webclient, and do not choose frontend. The modal proceeds and the daemon returns `invalid startTime format 0001-01-01 00:00:00 +0000 UTC` and success message :) 

**To Reproduce**
1. Go to webclient
2. Try to create event without selecting frontend
3. Check browser console

**Expected behavior**
Create event modal should not proceed without all required fields are selected. 
Daemon should return correct error message, instead of `invalid startTime format 0001-01-01 00:00:00 +0000 UTC` 
there should be `no frontend specified, please specify your frontend`. (For this specific bug). 
Error messages should be send correctly by the daemon, otherwise, debugging will be in pain. 

**Screenshots**

image

**Additional Information**

I already fixed the issue regarding to showing success message on webclient with this PR (https://github.com/aau-network-security/haaukins-webclient/pull/232). However correct message should still be sent by daemon.

mrtrkmn:
Closes #723 

This PR is fixing incorrect order of success and failure messages issue described in the issue #723 

Screenshots to the fix: 

From daemon side: 

false-msg-fix-daemon

From webclient side: 

fix-to-return-false-msg

### DIFF
--- a/daemon/event.go
+++ b/daemon/event.go
@@ -163,9 +163,17 @@ func (d *daemon) CreateEvent(ctx context.Context, req *pb.CreateEventRequest) (*
 			tags[i] = t
 		}
 		evtag, _ := store.NewTag(req.Tag)
-		finishTime, _ := time.Parse(dbTimeFormat, req.FinishTime)
-		startTime, _ := time.Parse(dbTimeFormat, req.StartTime)
-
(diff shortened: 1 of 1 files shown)


Mikkelhost:
Looks good you can merge :)

